PR #6, opened 2013-06-07
Rewrites the reset_simulator script as a bash script

aegolden:
We just keep having trouble with the AppleScript form of this
script.  So instead of using AppleScript to hit the simulator's
Reset Contents and Settings menu item we'll just blow away
everything the simulator stored in Application Support.


aegolden:
I'm going to close this one because it's superseded by #7.

PR #11, opened 2013-06-08
Aegolden/ios5 compat

aegolden:
Changes to address some ARC and UIAutomation bugs that exhibit in iOS 5 and earlier so that the Subliminal framework will work more reliably on these older systems.

In particular, these changes allow the example app's integration tests to work correctly in the iOS 5 simulator, and the `subliminal-test` CI script to target older versions of the SDK.

### DIFF
--- a/Integration Tests/SLIntegrationTestsAppDelegate.m
+++ b/Integration Tests/SLIntegrationTestsAppDelegate.m
@@ -82,7 +82,7 @@
 
 - (void)verifyTerminalConnectionAndRunTests {
     // Wait for UIAutomation to evaluate our command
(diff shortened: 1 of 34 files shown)


aegolden:
Just force pushed to update the two commits that I think we're ready to take (which are also sufficient for getting the example app working on iOS 5.1).  I've removed the last commit, which made the Subliminal unit tests "pass" on iOS 5.1 because I agree it's not a great approach.  I'm not sure what the right approach is, but I agree we shouldn't have to break the functionality of SLMainThreadRef (or registerTargetForAction, for that matter) in the general case just to make unit tests work on iOS 5.1.

PR #17, opened 2013-06-10
Aegolden/fix slassert true

aegolden:
Fixes SLAssertTrue per Max Gabriel's suggestion.


wearhere:
Thanks for the fix (and unit test :thumbsup:), @aegolden. Looked good, just added a small comment describing the purpose of the test.

PR #28, opened 2013-06-13
Add a set of "minimal instructions" to the README

aegolden:
Adds a set of minimal instructions, for developers who are
comfortable working with Xcode build settings and target
configurations and already have a project and integration
testing target that they want to use.


wearhere:
Hey @aegolden you and I talked offline about this a while back and agreed not to take it in favor of a more holistic reorganization (#82), but I never closed it out--doing so now. Thanks

Issue #33, opened 2013-06-22
Enable `subliminal-test` to run using Xcode 4.6 even when Xcode 5 is installed.


wearhere:
Installing Xcode 5 changes the Xcode developer directory which messes up `subliminal-test` insofar as Subliminal isn't fully compatible with the Xcode 5 toolchain.

wearhere:
This is a stopgap for #32.

Issue #31, opened 2013-06-21
Can't build workspace using subliminal-test


MaxGabriel:
I'm using standard CI script from the README to run Subliminal through Jenkins, except I'm using Cocoapods, so I need to choose a workspace to build:

```
# Run the tests in the non-retina iPhone Simulator
DEVICE="iPhone"

OUTPUT_DIR=test-reports
mkdir -p OUTPUT_DIR

# Returns 0 on success, 1 on failure
# Log output and screenshots will be placed in $OUTPUT_DIR
"ios-sdk/Heyzap/Pods/Subliminal/Supporting Files/CI/subliminal-test" -workspace "ios-sdk/Heyzap/Heyzap.xcworkspace" -sim_device "$DEVICE" -output "$OUTPUT_DIR" -login_password "hunter2"
```

But when I run the script I get the following error:

```
Building app...
Build settings from command line:
    SDKROOT = iphonesimulator6.1
    SYMROOT = /tmp/subliminal-test.build.gVh4uT
    TARGETED_DEVICE_FAMILY = 1
    VALIDATE_PRODUCT = NO

xcodebuild: error: '' is not a project file.


ERROR: Could not build application.
```

The script works fine against a standard `.xcodeproj` file.

MaxGabriel:
Ah, now that I check the `subliminal-test` script, it looks like it isn't using the workspace argument at all:

```
$BUILD_TOOL\
        -project "$PROJECT"\
        -scheme "$SCHEME"\
        -configuration "$CONFIGURATION"\
        -sdk "$SDK"\
        $TARGETED_DEVICE_FAMILY_ARG\
        SYMROOT="$BUILD_DIR"\
        VALIDATE_PRODUCT=NO\
        clean build
```

MaxGabriel:
Cool, so swapping out the `-project "$PROJECT"` argument with `-workspace "$WORKSPACE"` fixes it for me, though obviously there's still a bug in the script. I don't really know any shell scripting -- is there a good way to express that if `$PROJECT` is blank, use `-workspace "$WORKSPACE"`, else use `-project "$PROJECT"` ?

wearhere:
Good catch @MaxGabriel! Can you try building your workspace off of https://github.com/inkling/Subliminal/tree/jeff/respect_workspace_arg? It looks like `subliminal-test` will now properly prepare a workspace build command but I don't have any to build.

wearhere:
[...]

PR #34, opened 2013-06-22
`subliminal-test` can now be run alongside the Xcode 5 toolchain, and respects the "workspace" argument.

wearhere:


wearhere:
This fixes both #31 and #33 because I couldn't test the former without the latter.

wearhere:
For anyone keeping score, the attempt to convert `reset_simulator` into a text document (9f54bc83dc6d110dbe840cda5a35cbab60542d2b) [failed](https://travis-ci.org/inkling/Subliminal/builds/8333273#L1104) and so was removed. :cry:

wearhere:
Bloody hell, `reset_simulator` timed out again even as a compiled script. I'll look into replacing it with a programmatic solution tomorrow.

HaloZero:
@wearhere where was the script timing out? I felt like adding a delay between application launch and trying to set the menu option would work.

README.md
```diff
@@ -277,9 +277,22 @@ application and tests.
...
+Master, but there is at least [one serious bug](http://openradar.appspot.com/radar?id=3115412) 
+in the current 5.0 developer tools that would need to be resolved by Apple before we 
+could fully support that toolchain.
+
+If you have Xcode 5 installed, your [continuous integration script](#continuous-integration) 
+must select the Xcode 4.6 toolchain before invoking `subliminal-test`:
+
+```sh
+export DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer
+```
```
MaxGabriel, line 295: Is this different from `xcode-select`?

wearhere, line 295: It's documented in `xcode-select`'s `man` page as having the same effect as `xcode-select --switch <path>`--but only within the same shell & subshells, so devs can use the Xcode 5 toolchain normally if they like. The other benefit vs. `xcode-select --switch` is that setting `DEVELOPER_DIR` doesn't require superuser privileges.

MaxGabriel:
Yep, this works perfectly:

[x] On a workspace with only Xcode 4.6 installed
[x] On a workspace with both Xcode 4.6 and Xcode 5 installed
[x] On a regular project, with both Xcode 4.6 and Xcode 5 installed

wearhere:
@HaloZero It's difficult to say where the script was timing out because Applescript errors [don't include line-numbers](https://travis-ci.org/inkling/Subliminal/builds/8333273#L1104) (at least not in a form which corresponds to the decompiled source). Your guess is reasonable. I don't think that it's worth debugging when we can reset the content and settings directly (fbb4ddf), though--I'd rather not worry about timing issues across different environments (to say nothing of the fragility of trying to script applications' interfaces anyway).

@MaxGabriel Thanks for running those tests. @inkling/committers, this is ready for review.

wearhere:
Lol that's not the right group. And it appears that I don't have permissions to create a more appropriate one... ok how about @ahaneyinkling or @aegolden, one of you please.

PR #39, opened 2013-07-02
Add capturing screenshots to SLDevice

MaxGabriel:
Hey I have to run but I saw Jeff was on so I thought I'd put this up (Not ready to merge -- no documentation).
1. Any ideas on how to test screenshot functionality? Haven't thought about this too hard but presumably difficult.
2. How to represent the `Rect`, `Point`, etc. objects in UIAutomation? For now I just made a function that takes a `CGRect` and turns it into the literal string used to make the `Rect` object. Probably at the least I should move this to like an `SLGeometry` class.

### DIFF
--- a/Unit Tests/SLGeometryUnitTests.m
+++ b/Unit Tests/SLGeometryUnitTests.m
@@ -1,0 +1,32 @@
+//
+//  SLGeometryUnitTests.m
+//  Subliminal
+//
+//  Created by Maximilian Tagher on 7/3/13.
+//  Copyright (c) 2013 Inkling. All rights reserved.
+//
+
+#import <SenTestingKit/SenTestingKit.h>
+#import "SLGeometry.h"
+#import <Foundation/Foundation.h>
+#import <UIKit/UIKit.h>
+
+@interface SLGeometryUnitTests : SenTestCase
+
+@end
+
+@implementation SLGeometryUnitTests
+
+- (void)testSLUIARectFromCGRectThrowsWhenGivenCGRectNull
+{
+    STAssertThrowsSpecificNamed(SLUIARectFromCGRect(CGRectNull), NSException, NSInternalInconsistencyException, @"An internal inconsistency exception should be thrown if CGRectNull is passed");
+}
+
+- (void)testSLUIARectFromCGRectConvertsCorrectly
+{
+    STAssertEqualObjects(SLUIARectFromCGRect(CGRectMake(0.5f, -10, 20, 30)), @"{origin:{x:0.500000,y:-10.000000}, size:{width:20.000000, height:30.000000}}", @"A rect of {{0.5f,-10},{20,30}} should return the correct UIA representation");
+}
+
+
+
+@end
--- a/Integration Tests/Tests/SLGeometryTest.m
+++ b/Integration Tests/Tests/SLGeometryTest.m
@@ -1,0 +1,39 @@
+//
+//  SLGeometryTest.m
+//  Subliminal
+//
+//  Created by Maximilian Tagher on 7/3/13.
+//  Copyright (c) 2013 Inkling. All rights reserved.
+//
+
+#import <Subliminal/Subliminal.h>
+#import "SLIntegrationTest.h"
+#import "SLGeometry.h"
+#import "SLTerminal.h"
+
+@interface focus_SLGeometryTest : SLIntegrationTest
+
+@end
+
+@implementation focus_SLGeometryTest
+
++ (NSString *)testCaseViewControllerClassName {
+    return @"SLGeometryTestViewController";
+}
+
+- (void)testTheRectOfTheMainWindowIsEquivalent
+{
+    const CGRect UIARect = SLCGRectFromUIARect(@"UIATarget.localTarget().frontMostApp().navigationBar().rect()");
+    const CGRect UIKitRect = [SLAskApp(navigationBarFrameValue) CGRectValue];
+    
+    SLAssertTrue(CGRectEqualToRect(UIARect, UIKitRect), @"The frame of the main window should be the same when coming from UIAutomation or UIKit");
+}
+
+- (void)testTest
+{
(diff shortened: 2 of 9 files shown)


wearhere:
If UIAutomation can be directed to write the screenshot into the app's sandbox, then the app could operate on it to test-- @aegolden, thoughts? I recall you doing something similar for Inkling.

I think `SLUIARectFromCGRect` looks good, but yeah I think it would be a good idea to move it into an `SLGeometry` file. The inverse function (`Rect` -> `CGRect` string) is used by, and defined in, `-[SLUIAElement rect]`, and could probably be moved to `SLGeometry` as well--something like:

``` objc
// `UIARect` is some string which evaluates to a `Rect`
CGRect SLCGRectFromUIARect(NSString *UIARect) {
    NSString *CGRectString = [[SLTerminal sharedTerminal] evalFunctionWithName:@"SLCGRectStringFromJSRect"
                                                                        params:@[ @"rect" ]
                                                                          body:@"if (!rect) return '';\
                                                                                      else return '{{' + rect.origin.x + ',' + rect.origin.y + '},\
                                                                                                       {' + rect.size.width + ',' + rect.size.height + '}}';"
                                                                      withArgs:@[ UIARect ]];
    return ([CGRectString length] ? CGRectFromString(CGRectString) : CGRectNull);
}
```

MaxGabriel:
Ok, I think only testing and code review left.

Integration Tests/Tests/SLTextFieldTest.m
```diff
@@ -58,6 +58,9 @@ - (void)setUpTestCaseWithSelector:(SEL)testSelector {
 #pragma mark - SLTextField test cases
 
 - (void)testSetText {
+    [[SLDevice currentDevice] captureScreenshotWithFilename:@"aName"];
+    [[SLDevice currentDevice] captureScreenshotWithFilename:@"rectOnly" inRect:CGRectMake(100, 100, -100, -100)];
+    [[SLDevice currentDevice] captureScreenshotWithFilename:@"rectOnly" inRect:CGRectMake(0, 0, 100, 100)];
```
wearhere: Please remove these screenshot lines.

wearhere:
Thanks @MaxGabriel, some feedback for you.

wearhere:
@MaxGabriel, it looks like we left this request almost ready to go. It's a little hard to read now, so to summarize the remaining tasks:
- Remove ivar block [here](https://github.com/inkling/Subliminal/pull/39/files#L0R31)
- Remove `const` specifier [here](https://github.com/inkling/Subliminal/pull/39/files#L4R193)

And then, can you [fixup](https://help.github.com/articles/interactive-rebase) your amendments into the original commits? At least commits like 4ab415d, 5ff5e79, and 6b39feb, to ensure that the repo's healthy at every commit; and also feedback commits too if you can, to make the history easier to read.

MaxGabriel:
Yeah I'm on vacation with my extended family so I got distracted -- I think I can finish up on Sunday or Monday.

wearhere:
Oh sure @MaxGabriel, whenever you have a chance.

Unit Tests/SLGeometryUnitTests.m
```diff
@@ -0,0 +1,32 @@
...
+@implementation SLGeometryUnitTests
+
+- (void)testSLUIARectFromCGRectThrowsWhenGivenCGRectNull
+{
+    STAssertThrowsSpecificNamed(SLUIARectFromCGRect(CGRectNull), NSException, NSInternalInconsistencyException, @"An internal inconsistency exception should be thrown if CGRectNull is passed");
+}
+
+- (void)testSLUIARectFromCGRectConvertsCorrectly
+{
+    STAssertEqualObjects(SLUIARectFromCGRect(CGRectMake(0.5f, -10, 20, 30)), @"{origin:{x:0.500000,y:-10.000000}, size:{width:20.000000, height:30.000000}}", @"A rect of {{0.5f,-10},{20,30}} should return the correct UIA representation");
```
MaxGabriel, line 27: Specifying the number of zeroes feels really wrong to me, but so does modifying the implementation for the benefit of the test. Maybe I should just delete this test case?

MaxGabriel:
Ok, rebased. I had never rebased before, but in the future I'll try to do that before pushing. All of the unit and integration tests now pass for every commit in the PR.

Unit Tests/SLGeometryUnitTests.m
wearhere, line 27: I don't think that having to specify the number of zeroes is too problematic--testing a string formatting routine means guaranteeing a specific output.

I oftentimes feel that `%f`'s output is a little bit verbose, though. If you used `%g`, it would strip all "unnecessary" zeroes--`10.52` -> `10.52`, but `10.50` -> `10.5` and `10.0` -> `10`--and that might be a bit more common-sensical as far as reading and testing the output.

wearhere:
Thanks for rebasing @MaxGabriel.

> in the future I'll try to do that before pushing.

Cool. I'd been thinking about moving toward an [`xctool`-style workflow](https://github.com/facebook/xctool#workflow)--I think it will make pull requests a lot easier to manage.

This looks good except for that a first draft of the tests seems to have been included in 7cc0becc68176b7a9a080e573364b30485b7077b. Can you extract that and squash it into c2176c9ebcbe5aaa78abd127fdca7a54284aae73? You can do that by
1. Resetting 7cc0becc68176b7a9a080e573364b30485b7077b in a rebase (by marking it as `edit`, then when the rebase stops, executing `git reset HEAD^`).
2. Committing the documentation changes, then (separately) the testing changes.
3. Finishing the rebase.
4. Then doing a second rebase where you squash c2176c9ebcbe5aaa78abd127fdca7a54284aae73 into the testing changes commit.

phatmann:
+1 for this feature. I am currently using Zucchini, and I need screenshot compare if I am going to move over to Subliminal.

It has been 14 days since @MaxGabriel submitted his commit. Would you consider [squashing his commit for him](http://stackoverflow.com/questions/11396069/squash-to-only-one-proper-commit-for-github-pull-request) so we can close this pull request?

PR #56, opened 2013-08-13
Execute test classes in alphanumeric order

ibsh:
We have a few tests that need to be executed in a specific order at app launch, so I've switched use of sets to ordered arrays.

I'm happy to approach this differently if it's a feature you have any interest in merging.

### DIFF
--- a/Integration Tests/SLIntegrationTestsAppDelegate.m
+++ b/Integration Tests/SLIntegrationTestsAppDelegate.m
@@ -39,7 +39,7 @@
 @end
 
 @implementation SLIntegrationTestsAppDelegate {
-    NSSet *_tests;
+    NSArray *_tests;
     NSString *_terminalStartupResult;
 }
 
@@ -56,7 +56,7 @@
         // Filter the tests for the SLTestController
         // so that the SLTestsViewController only displays appropriate tests
         _tests = [SLTestController testsToRun:[SLTest allTests] withFocus:NULL];
-        rootViewController = [[SLTestsViewController alloc] initWithTests:[_tests allObjects]];
+        rootViewController = [[SLTestsViewController alloc] initWithTests:_tests];
     }
     UINavigationController *navController = [[UINavigationController alloc] initWithRootViewController:rootViewController];
     self.window.rootViewController = navController;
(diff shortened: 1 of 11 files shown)


wearhere:
Thanks for the request @ibsh. Can you describe your use case?

ibsh:
Sure thing; we have a registration screen or walkthrough (depending on device) which only appear at the first start of a newly installed app instance, and we need to test their behaviours. Once any other test has run, those screens are gone and untestable. We could delve into our codebase to build a handle to get back to them, but this seemed the simpler route!

wearhere:
You know we once had the notion of a "startup test" in Subliminal for just this, but we took it out because using an ordering made tests less flexible. For instance, how do you run a single test if you require an ordering--won't that test require that the registration/walkthrough test run too, beforehand? What if the registration/walkthrough test can't dismiss the registration/walkthrough--does the other test have to fail?

You've definitely identified a need for some common setup, but I'd suggest that you factor that setup into a `-setUpTest` implementation shared by all your tests (that require that setup)--that is, define an [abstract test](http://inkling.github.io/Subliminal/Documentation/Classes/SLTest.html#//api/name/isAbstract) from which your other tests derive. This way, each test will establish its own state and can be run in isolation. And if you have the abstract test's `-setUpTest` do its work through an [app hook](https://github.com/inkling/Subliminal/blob/master/README.md#manipulate-the-application-directly), then changes in the registration/walkthrough UI won't necessarily cause your other tests to fail. You can have the registration/walkthrough test override this set-up to actually test that UI, but the other tests can effectively skip it.

ibsh:
I agree with you in principle on all your points, but unfortunately I have to compromise on principle in this instance!

Actually, all our other tests are fine without the startup test, as skipping the startup screens as part of common setup (which we've already employed as part of an app navigation class) is pretty trivial. So we don't have an issue with flexibility in this case.

So it's no longer common setup, just a (relatively low priority) test case in and of itself. And I've chatted with the devs (I'm in QA) about an app hook for returning to these elements, but it's a bad cost/benefit equation in this case.

I understand and appreciate that you guys prefer a "principled architecture", however =)

wearhere:
Oh, so it sounds like your problem isn't dismissing the registration/walkthrough UI--it's getting back to it, for the registration/walkthrough test itself. Yeah I can understand why that might be a tricky problem to solve. Thanks for understanding our position though! Closing.

PR #67, opened 2013-08-16
Reset the simulator and rebuild the app before restarting instruments.

wearhere:
Delaying for a longish period of time before restarting (f7a0ff565eab437d04d145aabf44798354422fdd)
did not work: https://travis-ci.org/inkling/Subliminal/builds/10263900#L2210.
Perhaps `instruments` is rejecting the build for some reason.
Certainly, when we have seen a similar error locally, it is because the build
was invalid (for instance, it was built for the device but attempted to be run
in the simulator).

### DIFF
--- a/Supporting Files/CI/subliminal-test
+++ b/Supporting Files/CI/subliminal-test
@@ -169,7 +169,7 @@ case $1 in
 		shift 1;;
 
     --quiet_build)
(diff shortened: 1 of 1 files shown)


wearhere:
Yessss https://travis-ci.org/inkling/Subliminal/builds/10265765#L2205 . We may have a winner.

PR #75, opened 2013-08-27
Subliminal fully supports running tests on devices.

wearhere:
`instruments` will profile an application on a device only if that application is already installed on the device. This could and did run unnoticed if you already had the application you were testing installed on your device i.e. from Xcode. Now Subliminal (re)installs the application under test on the target device by using Apple's `MobileDevice` framework.

### DIFF
--- a/Supporting Files/CI/device-install.h
+++ b/Supporting Files/CI/device-install.h
@@ -1,0 +1,265 @@
+//
+//  device-install.h
+//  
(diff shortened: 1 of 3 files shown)


wearhere:
Rebased to pick up bb53b6b8f10bde50dab985a012c2d62d8daab392, which will fix the last build's failure.

wearhere:
Updated, @aegolden.

PR #80, opened 2013-09-05
Test order is now fully randomized. (closes #78)

### LINKED ISSUE #78: Reproducing test runs.
### Goals

To track down [test pollution](http://pivotallabs.com/find-test-pollution-rspec/)

> seemingly false negatives or false positives in a test suite[, occurring] when some shared state is unintentionally modified, or unintentionally read and used in a test

it would be useful to:
1. be able to reproduce failing test runs in the order that the tests ran.
2. (within that run) skip to a particular test--e.g. the last one before the tests started failing, on the theory that that test was the polluter.
### Proposal

Tests are currently run in a pseudorandom order (the order that the runtime discovers them, which I'm pretty sure is the order in which they're linked). This order may be reused across runs but not necessarily.

I propose that we:
- fully randomize test ordering
- if one or more tests fail, log the seed that was used to randomize that run
- allow the tests to be rerun using that seed, thus in a consistent order

This functionality would be exposed to the user via a new API on `SLTestController`:

``` objc
- (void)runTests:(NSSet *)tests usingSeed:(unsigned int)seed withCompletionBlock:(void (^)(void))completionBlock;
```

> Note: `-[NSSet allObjects]` does not return objects in a determinate order, but we would need a consistent basis for randomization. So, I propose that the test controller would alphabetize `[tests allObjects]` before randomizing.

To accomplish goal 2, given a particular set of tests, seed, and test

``` objc
- (void)runTests:(NSSet *)tests usingSeed:(unsigned int)seed skippingToTest:(Class)firstTest withCompletionBlock:(void (^)(void))completionBlock;
```

 the test controller would iterate through the tests, using the predefined order, until `firstTest` was reached, then would begin executing the tests.
### Questions
1. Do we even need both APIs, or would the second suffice?
2. Should these APIs be `DEBUG`-only to prevent them being used in CI environments?

wearhere:
Tests are now sorted and then randomized prior to being run.
If one or more tests fail, the seed used to randomize the order will be logged.
That seed may then be used to rerun the tests in the same order.

If the seed was predetermined (by the user), a warning will be logged
and `subliminal-test` will fail the run (under the assumption that tests should
generally be run randomized, to prevent test pollution, and that a seed will
be predetermined only for debugging purposes.)

Test relative order will be maintained regardless of focus.

### DIFF
--- a/Sources/Classes/SLTestController.m
+++ b/Sources/Classes/SLTestController.m
@@ -35,6 +35,8 @@
 #import <objc/runtime.h>
 
 
+const unsigned int SLTestControllerRandomSeed = UINT_MAX;
+
 static NSUncaughtExceptionHandler *appsUncaughtExceptionHandler = NULL;
 static const NSTimeInterval kDefaultTimeout = 5.0;
 
@@ -90,8 +92,9 @@ static void SLUncaughtExceptionHandler(NSException *exception)
 
 @implementation SLTestController {
     dispatch_queue_t _runQueue;
-    BOOL _runningWithFocus;
-    NSSet *_testsToRun;
+    unsigned int _runSeed;
+    BOOL _runningWithFocus, _runningWithPredeterminedSeed;
+    NSArray *_testsToRun;
     NSUInteger _numTestsExecuted, _numTestsFailed;
     void(^_completionBlock)(void);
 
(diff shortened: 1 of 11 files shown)


Sources/Classes/SLTestController.m
```diff
@@ -118,21 +132,78 @@ + (instancetype)sharedTestController {
...
+    for (i = 0; i < sizeof now; i++) {
+        seed = seed * ( UCHAR_MAX + 2U ) + p[i];
+    }
+    
+    return seed;
+}
+
+/// a condensed version of the `uniform_deviate` function
+/// also from http://eternallyconfuzzled.com/arts/jsw_art_rand.aspx
+u_int32_t random_uniform(u_int32_t upperBound) {
```
wearhere, line 159: @aegolden I ended up using this instead of just taking the modulus because I was getting some outliers in `-[SLTestControllerTests testTestsRunInRandomOrderWhen_SLTestControllerRandomSeed_IsSpecified]`--this did actually seem to produce a more uniform distribution.

wearhere:
In order to fix the failure in `-[SLTestControllerTests testTestsRunInRandomOrderWhen_SLTestControllerRandomSeed_IsSpecified]`, we need to decide on a better standard for randomness and/or improve our randomization algorithm to meet that standard. Over repeated invocations of the test case, with 100 trials per case, every permutation of the tests is consistently observed at least once; but as far as the distribution of the orders--it's all over the place. Even a tolerance for **90%** deviations from the expected count ([here](https://github.com/inkling/Subliminal/pull/80/files#L5R108)) occasionally fails to cut it. I'd like to talk about this with you when you have a chance, @aegolden.

wearhere:
Thanks for the new test @golden. That's so much more robust than the old one.

PR #85, opened 2013-09-27
Adds SLSwitch

j-mutter:
This adds an `SLSwitch` class which enables wraps `UIASwitch`'s `setValue` method, and provides a property to access the current value of the switch as a bool instead of a string.

### DIFF
--- a/Sources/Classes/UIAutomation/User Interface Elements/SLSwitch.h
+++ b/Sources/Classes/UIAutomation/User Interface Elements/SLSwitch.h
@@ -1,0 +1,20 @@
+//
+//  SLSwitch.h
+//  Subliminal
+//
+//  Created by Justin Mutter on 2013-09-13.
+//  Copyright (c) 2013 Inkling. All rights reserved.
+//
+
+#import <Subliminal/Subliminal.h>
+
+#define ON YES
+#define OFF NO
+
+@interface SLSwitch : SLButton
+
+@property (readonly,getter=isOn) BOOL on;
+
+- (void)setValue:(BOOL)value;
+
+@end
--- a/Sources/Classes/UIAutomation/User Interface Elements/SLSwitch.h
+++ b/Sources/Classes/UIAutomation/User Interface Elements/SLSwitch.h
@@ -9,7 +9,7 @@
 #import "SLButton.h"
 
 /**
- `SLSwitch` will match against any `UIButton` object, but provides extra functionality for instances of `UISwitch`.
+ `SLSwitch` matches against instances of `UISwitch`
  
   Tappipng an `SLSwitch` will toggle its value, which can be queried with the `on` property, or the switch may be set to a specific value.
  */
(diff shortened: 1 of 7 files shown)


Sources/Classes/UIAutomation/User Interface Elements/SLSwitch.h
```diff
@@ -0,0 +1,33 @@
...
+//  Subliminal
+//
+//  Created by Justin Mutter on 2013-09-13.
+//  Copyright (c) 2013 Inkling. All rights reserved.
+//
+
+#import "SLButton.h"
+
+#define ON YES
+#define OFF NO
```
wearhere: I'm loathe to dump symbols with names as generic as "ON" and "OFF" into the global namespace. At the same time, it is nice to layer "off/on" semantics onto `value`. What do you think about getting rid of `setValue:` and just making `on` readwrite? `setOn:YES` would be pretty clear imo--that's how `UISwitch` is set up, actually.

Sources/Classes/UIAutomation/User Interface Elements/SLSwitch.h
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ `SLSwitch` will match against any `UIButton` object, but provides extra functionality for instances of `UISwitch`.
```
wearhere: I think we should restrict `SLSwitch` to match _only_ instances of `UISwitch`:

(in `SLSwitch.m`):

``` objc
- (BOOL)matchesObject:(NSObject *)object {
    return ([super matchesObject:object] && [object isKindOfClass:[UISwitch class]]);
}
```

There's no guarantee that `value` will be boolean, otherwise. Also, this best preserves parity with UIAutomation, where `UIASwitch` matches instances of `UISwitch`.

Sources/Classes/UIAutomation/User Interface Elements/SLSwitch.h
```diff
@@ -0,0 +1,33 @@
...
+ */
+@interface SLSwitch : SLButton
+
+/**
+ The boolean value of the switch.
+ */
+@property (readonly,getter=isOn) BOOL on;
+
+/**
+ Sets the value of the switch, regardless of it's current value.
```
wearhere: Would you mind fixing the typos here, please? "Tappipng", "it's", "querried".

wearhere:
Hey thanks for the request @j-mutter. I really appreciate you including documentation and tests. I've made a few comments. Also, Travis doesn't yet have Xcode 5 installed, unfortunately, so the build will fail like [so](https://travis-ci.org/inkling/Subliminal/builds/11852137#L5431) unless you set `SLSwitchTestViewController.xib` to "open in Xcode 4.6" in Xcode's right sidebar:

screen shot 2013-09-30 at 4 38 51 pm

Thanks!

j-mutter:
@wearhere Done. I had been using `setValue` instead of `setOn` because that's how it's handled in `UIASwitch`, but matching `UISwitch` also makes sense. And not sure how I completely spaced on the lack of `matchesObject:`, but that's sorted as well.

wearhere:
Looks great! Thanks again.

PR #89, opened 2013-10-12
Add `SLEditingMenu` and `SLEditingMenuItem`.

wearhere:
To allow the manipulation of the application's editing menu and the items it contains.

It makes me frustrated that the editing menu intermittently appears to fail on Travis. I still think it's worth adding to the framework, though, since it works in other environments.


dhchow:
Would love to see this get in as well. Can an exception be made for these implementations since Travis is what's preventing the tests from passing?

PR #95, opened 2013-10-21
Allows various text views to handle "tap-hold" characters (fixes #91)

### LINKED ISSUE #91: SLKeyboard's typeString can't handle character variants that require a tap-hold
SLKeyboard's typeString method is able to change key-planes as needed, but seems to be incapable of handling key variants that require a tap-hold on the default key variant.

aegolden:
Adds an internal variant of SLKeyboard's typeString method,
which can fall back on the JavaScript setValue method if
the normal type string mechanism throws an exception.

Causes Subliminal UI element classes to use the new method
when appropriate.


wearhere:
Ok, some comments on documentation @aegolden, thanks.

aegolden:
Updated, @wearhere.

PR #106, opened 2013-12-11
Add NS_FORMAT_FUNCTION declarations to SLComposeString and SLComposeStringV.

ewanmellor:
This means that we get warned if the arguments do not match the
format specifiers in the failure descriptions for SLAssert*.


wearhere:
This looks like a good change--thanks @ewanmellor, and sorry for the delay in reviewing. I've just fixed a couple of things in the build process; would you mind rebasing this request on top of master so the tests will run before merging?

ewanmellor:
Rebased under PR #111.

PR #107, opened 2013-12-18
Pass "numTestsExpected" to logTestingFinish

aegolden:
Adds a new parameter to the logTestingFinished method for the
number of tests that were expected to run, and causes the method
to log "Testing did not finish..." if the number of tests that
executed is not equal to the number of tests that were expected.


aegolden:
Actually let's not do this.  I don't know of any reasons for the tests executed to not equal the tests expected.  I was looking at jobs where expected test _cases_ didn't run due to failures in various setUpTest methods, but these changes don't help tracking executed vs. expected test cases.

PR #111, opened 2014-01-15
Add NS_FORMAT_FUNCTION declarations to SLComposeString and SLComposeStringv.

ewanmellor:
This means that we get warned if the arguments do not match the
format specifiers in the failure descriptions for SLAssert*.


wearhere:
Thanks @ewanmellor! I think there was a timing issue in one of the integration tests, but that's most certainly unrelated to your changes--the rest, in particular the unit tests, checked out. Merging.

PR #115, opened 2014-01-23
Update Subliminal documentation

j2bbayle:
A few updates to [README.md](https://github.com/inkling/Subliminal/blob/master/README.md):
- Subliminal no longer has issues with Xcode 5.0.1;
- Swiping table view cells doesn't work with the iOS 7 Simulator.

### DIFF
--- a/README.md
+++ b/README.md
@@ -530,6 +530,12 @@ limitations of Apple's frameworks or bugs therein. Other issues are tracked
 *	UIAutomation cannot drag scroll views when running in the iOS 7 Simulator.
 
 	`SLElement` implements a workaround.
+	
+	This issue also affects swiping table view cells since they [now internally use a
+	scroll view](http://www.curiousfind.com/blog/646) to implement the 'swipe-to-delete' 
+	behavior. Using drag gestures will tap the cell but **not** show the deletion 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -534,6 +530,11 @@ limitations of Apple's frameworks or bugs therein. Other issues are tracked
 *	UIAutomation cannot drag scroll views when running in the iOS 7 Simulator.
 
 	`SLElement` implements a workaround.
+	
+	This issue also affects swiping table view cells. Using drag gestures will tap tap
```
wearhere: I think it's appropriate to put the "more info" link here because the link supports why it is that the issue affects swiping, not why Subliminal can't work around the issue. What if you said something like "affects swiping table view cells because they [use scroll views on iOS 7](http://www.curiousfind.com/blog/646)"?

wearhere: "tap tap"

wearhere:
Cancelled the build because these are just documentation changes. Please tag both commits with "[skip ci]" when you rebase, @j2bbayle.

j2bbayle:
Sorry about that @wearhere. That should be better now!

wearhere:
Very lucid comments @j2bbayle. Thanks!

PR #116, opened 2014-01-28
Get the path for the deepest matching element

j2bbayle:
We usually want to grab the path for the deepest (leaf) matching object in the tree; this is especially true in iOS 7 since Apple changed the accessibility hierarchy and inserted intermediate views or accessibility elements in between.

Let’s modify `-rawAccessibilityPathToElement:favoringSubviews:` to search for the deepest matching object instead of return the first the method encounters.


wearhere:
For reference, a case where Subliminal previously failed was in matching links within web views. `UIWebView` generates a hierarchy of several accessibility elements with the same or similar accessibility information, and from UIAutomation's perspective all of them are valid, but only the leaf element is visible and tappable.

wearhere:
Elegant fix. Looks good as soon as Travis finishes.

PR #120, opened 2014-02-05
Increase timeout duration for Instruments to launch

j2bbayle:
Subliminal script originally waits 30 minutes when launching Instruments and the app. If nothing has been written after these 30 seconds, the Simulator is killed and it tries it all again — if the retry count limit has not yet been reached.

In older machines, 30 seconds may not be enough for Instruments and the app to launch. Let’s implement a sort of exponential fall back and increase the timeout duration after each try: 30 seconds, 45 seconds, 1 minutes and then 2 minutes.


Supporting Files/CI/subliminal-test
```diff
@@ -614,10 +621,11 @@ launch_instruments () {
 	return $instruments_exit_status
 }
 
-(( MAX_INSTRUMENTS_RETRY_COUNT = 3 ))
+INSTRUMENTS_LAUNCH_TIMEOUT=( 30 45 60 120 )
```
wearhere: As an array, can you pluralize this variable please?

wearhere:
Two small comments @j2bbayle, otherwise looks good! Elegant implementation.

j2bbayle:
Ready to be merged @wearhere -- although the test does not appear here even if it has been triggered in [Travis](https://travis-ci.org/inkling/Subliminal/builds/18313724).

PR #122, opened 2014-02-12
Adds `runTestsSync:` methods

j-mutter:
As promised, this adds the `runTestsSync:` methods, which are essentially just the `runTests:` methods without the dispatch, and they require a `finishTesting` argument to signal the final batch of tests.

### DIFF
--- a/Sources/Classes/SLTestController.h
+++ b/Sources/Classes/SLTestController.h
@@ -92,6 +92,9 @@
  will be executed on the main queue. The test controller will then signal 
  UIAutomation to finish executing commands.
  
+ Tests are run via the `runTestsSync:usingSeed:finishTesting:withCompletionBlock:` with `finishTesting:YES` but are
+ dispatched to a run queue.
+ 
  @param tests           The set of tests to run.
  @param seed            The seed to use to randomize the tests.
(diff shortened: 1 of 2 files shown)


j-mutter:
@wearhere Any thoughts on this yet?

j-mutter:
I hate to nag, but, uh, bump? @wearhere

wearhere:
Sorry @j-mutter, you're quite right to bump this; I've been delinquent.

I spent some time mulling this over 'cause it seems to be less coherent than I thought (or rather, didn't think--I kinda tossed the idea out there and didn't follow up with you about it). Most problematic, if I'm not mistaken each call to `runTestsSync:` causes the tests to "begin" again, including the associated log messages being re-emitted and the seed being re-chosen--but only the last seed chosen is logged if any tests fail.

Now, it might be possible to add additional parameters to the API to control this, but I thought that maybe it would be better for Subliminal to handle the segmentation internally and avoid leaking the test controller's implementation, so to speak. What do you think of #162 as a means to this end? I feel like I misled you with regard to the `runTestsSync` idea, but hopefully we can put something together to address your situation.

Once again, sorry for not responding to you sooner. I'll have limited internet access for the next few days but I'll try to respond as able.

wearhere:
Closing this now that #162's been merged.

PR #124, opened 2014-02-18
Use public github clone url in README.md

jazzychad:
I got bit a couple times copy/pasting the README commands in terminal by the private clone url, so here is a pull request to use the public clone urls instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ screencast [here](https://vimeo.com/67771344))
 Running the Example App
 -----------------------
 
-1. Clone the Subliminal repo: `git clone git@github.com:inkling/Subliminal.git`.
+1. Clone the Subliminal repo: `git clone https://github.com/inkling/Subliminal.git`.
 2. `cd` into the directory: `cd Subliminal`.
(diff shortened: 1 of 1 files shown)


wearhere:
Thanks @jazzychad. If I read [this GitHub guide](https://help.github.com/articles/which-remote-url-should-i-use) correctly, it's not that the "git@github.com" URL is private, it's that you don't have an SSH keypair attached to your GitHub account--but I think it's still a good idea to switch to the HTTPS URLs so that we don't require users to generate those keypairs. That said, if my guess about your situation is correct, could you amend the commit message accordingly and then force push to your branch? You can also put ["[skip ci]"](http://docs.travis-ci.com/user/how-to-skip-a-build/) in the body of the commit so that we don't have to wait for Travis to finish.

jazzychad:
sorry, yeah I guess I meant read/write where I said "private" - also you're correct, I didn't have keys setup on my particular machine I was playing with. Regardless, I'm glad you want to switch to https endpoint. I'll update accordingly.

jazzychad:
Ok, updated and pushed.

wearhere:
Great thanks!

PR #125, opened 2014-02-19
Guard against nil case for accessibilityElementAtIndex:

jazzychad:
If the accessibility protocol methods are not properly implemented (as
in custom controls for example), we might have a case where
accessibilityElementAtIndex: will return nil. This causes a crash in
slChildAccessibilityElementsFavoringSubviews: with a nil insertion
into an array. The fix is to guard against adding nil element to the
children array.

This is a proposed solution, if it is agreeable with your coding style. I tried to do the logWarning: inline but it caused Instruments to hang, but adding a dispatch_async solved it...

This problem was actually biting us, and it caused Instruments to end abruptly with an unhandled exception. Avoiding this nil insertion has cleared the way for our tests. (and also now we know we have some clean-up work to do on our custom controls regarding our accessibility implementation!)

### DIFF
--- a/Sources/Classes/UIAutomation/User Interface Elements/NSObject+SLAccessibilityHierarchy.m
+++ b/Sources/Classes/UIAutomation/User Interface Elements/NSObject+SLAccessibilityHierarchy.m
@@ -109,6 +109,13 @@
         if (count != NSNotFound && count > 0) {
             for (NSInteger i = 0; i < count; i++) {
                 id element = [self accessibilityElementAtIndex:i];
+                if (!element) {
+                    dispatch_async([[SLLogger sharedLogger] loggingQueue], ^{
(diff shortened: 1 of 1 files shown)


jazzychad:
:clap: yay Travis passed this time!

wearhere:
:+1: Dispatching async to the `loggingQueue` is exactly the way to go there--`SLLogger` isn't meant to be used from the main thread, apart from `SLLogAsync`.

As far as Travis goes, we've been seeing some weird issues on iOS 7 where UIAutomation taps the wrong keyboard key. The failures are intermittent--the way to fix it would probably be just to retry the `SLKeyboardTest` cases internally--and I just poke Travis to rerun the build when the cases fail. It is definitely nice when Travis passes first time, though.

PR #127, opened 2014-02-20
Add test tagging support to SLTest

jazzychad:
Add support for tagging tests and selectively running tests based on
tags. New +[SLTest tags] and +[SLTest testsWithTags:] methods added
for implementation. Updated unit tests included.

### DIFF
--- a/Sources/Classes/SLTest.h
+++ b/Sources/Classes/SLTest.h
@@ -54,6 +54,19 @@
 + (NSSet *)allTests;
 
 /**
+ Returns tests linked against the current target which contains a tag matching at least one tag specified in `tags`
+
+ This method will include a test in the returned set if any of the specified tags in the `tags` parameter match at least one of the tags in `+[SLTest tags]`. This is essentially a set union operation (i.e. specifying multiple tags will return all tests that match any of them).
+
+ This method may be useful to specify a certain set of tests in your app that pertain to a specific funtionality or to divide up your tests into separate test suites.
+
+ @param tags An array of tags to match against the tags specified in `+[SLTest tags]`
+
+ @return All tests (`SLTest` subclasses) linked against the current target which contains a tag matching at least one tag specified in `tags`
+ */
++ (NSSet *)testsWithTags:(NSArray *)tags;
+
+/**
  Returns the `SLTest` subclass with the specified name.
  
  This method may be used to retrieve a single `SLTest`, e.g. to pass to 
@@ -149,6 +162,19 @@
  @see -[SLTestController runTests:withCompletionBlock:]
  */
 + (BOOL)isFocused;
+
+/**
+ An array of NSStrings which represent tags that are associated with this test case.
+
+ Used by `+[SLTest testsWithTags:]` to determine inclusion in the test set.
+
+ Tags are case sensitive.
+
+ @return array of tags for this test, defaults to `nil` to indicate no tags.
+
+ @see -[SLTestController runTests:withCompletionBlock:]
+ */
++ (NSArray *)tags;
 
 #pragma mark - Running a Test
 /// ----------------------------------------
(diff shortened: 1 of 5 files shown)


jazzychad:
Nooooooo... how do we re-kick the Travis build?

edit: ok, looks like it was re-kicked?

wearhere:
Yup, I restarted the job. I don't know if you can do that--I think you might have to be an admin on the repo to see the "redo" icon at the top-right of a failed build. No problem though. :)

I can definitely see uses for running subsets of tests. I'm not sold on adding this particular mechanism to Subliminal, however--I think it might be too specific. What if you wanted to do a logical `AND` of tags, for instance?

Here are two more general mechanisms for running subsets of tests in production, and how they could be used to easily create `+testsWithTags:`. I apologize for the length of this post--I want to make clear that Subliminal can support this use case, just in a different way.

---

Both mechanisms presume that you create an [abstract](https://github.com/inkling/Subliminal/blob/master/Sources/Classes/SLTest.h#L87) subclass of `SLTest`:

``` objc
@interface MyAppTest : SLTest
+ (NSSet *)tags;
@end

@implementation
// to be overridden by subclasses
+ (NSSet *)tags { return nil; }
@end
```

and then use that abstract class as the basis for all your application's tests. Let's also assume that you specify the set of tags using an environment variable like so: `TAGS=one,two,three`. Then, the mechanisms:
1. **filter the set you pass to `+[SLTestController runTests:...]`** in response to criteria defined by the test classes:
   
   ``` objc
   NSSet *testsToRun = [SLTest allTests];
   if (getenv("TAGS")) {
       NSSet *tags = [NSSet setWithArray:[@(getenv("TAGS")) componentsSeparatedByString:@","]];
       testsToRun = [testsToRun filteredSetUsingPredicate:[NSPredicate predicateWithFormat:@"ANY SELF.tags IN %@", tags]];
   }
   [[SLTestController sharedTestController] runTests:testsToRun...];
   ```
2. **alter your tests' response to `+[SLTest supportsCurrentPlatform]`:**
   
   ``` objc
   @implementation MyAppTest
   + (BOOL)supportsCurrentPlatform {
       BOOL supportsCurrentPlatform = [super supportsCurrentPlatform];
       if (getenv("TAGS")) {
           NSSet *tags = [NSSet setWithArray:[@(getenv("TAGS")) componentsSeparatedByString:@","]];
           supportsCurrentPlatform = supportsCurrentPlatform && [[self tags] intersectsSet:tags];
       }
       return supportsCurrentPlatform;
   }
   @end
   ```

Rather than use tags to group tests, you could also make groups by using additional abstract classes. For instance, all of Inkling's tests that download books have the following superclass chain `ContentDownloadTest : AbstractInklingTest : SLTest`. When debugging, we can run all book-downloading tests by merely focusing `ContentDownloadTest`: subclasses inherit focus. And when in production, we could make `+isKindOfClass:[ContentDownloadTest class]` the basis of either of the mechanisms above.

**TL;DR** I'm not sure that the tags mechanism is sufficiently general or powerful to warrant addition to Subliminal core--but the scheme you describe could be layered on top of Subliminal using [abstract test classes](https://github.com/inkling/Subliminal/blob/master/Sources/Classes/SLTest.h#L87).

jazzychad:
Hey Jeff,

Thanks for the very thoughtful and thorough response.

That seems completely reasonable to me. We're already starting to have specialized base classes for our tests, so this mechanism would fit perfectly into what we are already starting to do.

I'll have a think about it for a while longer, but I'm pretty convinced this will work fine. Thanks for the feedback. I leave it up to you to close this request pending your final decision.

wearhere:
Ok thanks Chad! I'm glad you put up the proof of concept, it helped me clarify my thoughts around this. I'm open to revisiting this if you think of additional comments or additional users voice a demand for such a convenience to be built into Subliminal itself, but I'm going to close the request for now.

wearhere:
@jazzychad, since your request, I did indeed hear from a number of other users that they would like something similar. :) I've incorporated elements of your approach into #233 and would love to know what you think.

PR #129, opened 2014-02-20
Adds quick screenshot method to all SLElement objects

jzucker2:
Reduces steps and allows for easy screenshots of individual
elements. Should be useful in multiple runs for small UI changes.

### DIFF
--- a/Sources/Classes/UIAutomation/User Interface Elements/SLElement.h
+++ b/Sources/Classes/UIAutomation/User Interface Elements/SLElement.h
@@ -170,29 +170,6 @@
  */
 - (void)logElement;
 
-#pragma mark - Screenshots
-/// ----------------------------------------
-/// @name Screenshot an element
-/// ----------------------------------------
-
-/**
- Takes a screenshot of the specified element.
-
- The rect for the screenshot of the element is automatically provided.
-
- The image is viewable from the UIAutomation debug log in Instruments.
-
- When running `subliminal-test` from the command line, the images are also saved
- as PNGs within the specified output directory.
-
- @param filename A string to use as the name for the resultant image file.
-
- @exception `NSInternalInconsistencyException` if `rect` is `CGRectNull` for specified element.
-
- */
-
-- (void) screenshotWithFilename:(NSString *)filename;
-
 @end
 
 
--- a/Sources/Classes/UIAutomation/User Interface Elements/SLElement.h
+++ b/Sources/Classes/UIAutomation/User Interface Elements/SLElement.h
@@ -170,6 +170,29 @@
  */
 - (void)logElement;
 
+#pragma mark - Screenshots
+/// ----------------------------------------
+/// @name Screenshot an element
+/// ----------------------------------------
+
+/**
+ Takes a screenshot of the specified element.
+
+ The rect for the screenshot of the element is automatically provided.
+
+ The image is viewable from the UIAutomation debug log in Instruments.
+
+ When running `subliminal-test` from the command line, the images are also saved
+ as PNGs within the specified output directory.
+
+ @param filename A string to use as the name for the resultant image file.
+
+ @exception `NSInternalInconsistencyException` if `rect` is `CGRectNull` for specified element.
+
+ */
+
+- (void) screenshotWithFilename:(NSString *)filename;
+
 @end
 
 
--- a/Sources/Classes/UIAutomation/User Interface Elements/SLElement.m
+++ b/Sources/Classes/UIAutomation/User Interface Elements/SLElement.m
@@ -28,6 +28,7 @@
 #import "NSObject+SLAccessibilityDescription.h"
 #import "UIScrollView+SLProgrammaticScrolling.h"
 
+
 // The real value (set in `+load`) is not a compile-time constant,
 // so we provide a placeholder here.
 UIAccessibilityTraits SLUIAccessibilityTraitAny = 0;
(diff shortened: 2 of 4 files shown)


jzucker2:
My doc generation comments might not be good. Haven't worked that much for Obj-C. Also open to putting a timestamp if a filename isn't provided. I left notes in the comments in the patch.

Sources/Classes/UIAutomation/User Interface Elements/SLElement.h
```diff
@@ -170,6 +170,29 @@
...
 
+#pragma mark - Screenshots
+/// ----------------------------------------
+/// @name Screenshot an element
+/// ----------------------------------------
+
+/**
+ Takes a screenshot of the specified element.
+
+ The rect for the screenshot of the element is automatically provided.
```
wearhere: I don't think we need to mention this--it's kinda an implementation detail.

Sources/Classes/UIAutomation/User Interface Elements/SLElement.h
```diff
@@ -170,6 +170,29 @@
...
+
+ The image is viewable from the UIAutomation debug log in Instruments.
+
+ When running `subliminal-test` from the command line, the images are also saved
+ as PNGs within the specified output directory.
+
+ @param filename A string to use as the name for the resultant image file.
```
wearhere: Please note that this is optional.

Sources/Classes/UIAutomation/User Interface Elements/SLElement.h
```diff
@@ -170,6 +170,29 @@
...
+
+ @exception `NSInternalInconsistencyException` if `rect` is `CGRectNull` for specified element.
```
wearhere: This method abstracts away the use of `-[SLDevice captureScreenshot:inRect:]`, so I don't think it should leak the requirement of that method that _rect_ be non-null. Rather, if _rect_ is null, let's just have the method log a warning (`-[SLLogger logWarning:]`) like "Could not take screenshot with filename _filename_: Could not determine element's position on-screen."

Since `-rect` requires that the element be valid, though, we should include the standard `SLUIAElementInvalidException` warning here.

Sources/Classes/UIAutomation/User Interface Elements/SLElement.h
```diff
@@ -170,6 +170,29 @@
...
+
+ */
+
+- (void) screenshotWithFilename:(NSString *)filename;
```
wearhere: Can you name the method `captureScreenshotWithFilename:`, please? `screenshotWithFilename:` sounds like it might return an image.

Also, minor point of style: please remove the space between `(void)` and the method name.

Sources/Classes/UIAutomation/User Interface Elements/SLElement.m
```diff
@@ -398,4 +399,16 @@ - (void)logElementTree {
     [super logElementTree];
 }
 
+#pragma mark -
+
+- (void) screenshotWithFilename:(NSString *)filename
+{
+    // maybe consider using a timestamp? not sure if it's worth introducing the complexity of NSDateFormatter
```
wearhere: Yeah I don't think it's worth appending a timestamp. If the user really cares about distinguishing these screenshots, they can pass a filename.

wearhere:
Thanks for the PR @jzucker2! I think this is a solid addition. Besides the points of style I mention in the diff, one other thing to change is that all user interface elements can make use of this method, so it should be defined in `SLUIAElement.h/.m`.

jzucker2:
Updated as per your suggestions. Hope I understand you properly. Looking forward to hearing your feedback.

jzucker2:
Any idea why Travis CI would fail? This isn't drastically different than the previous commit. Seems to be an issue with iPad??

Sources/Classes/UIAutomation/User Interface Elements/SLUIAElement.h
```diff
@@ -399,6 +399,27 @@
...
+
+/**
+ Takes a screenshot of the specified element.
+
+ The image is viewable from the UIAutomation debug log in Instruments.
+
+ When running `subliminal-test` from the command line, the images are also saved
+ as PNGs within the specified output directory.
+
+ @param filename An optional string to use as the name for the resultant image file (provide 'nil' if you do not want the screenshot to have a name).
```
wearhere: I think it might be a bit clearer to say "provide `nil` to use a generic name."

Sources/Classes/UIAutomation/User Interface Elements/SLUIAElement.h
```diff
@@ -399,6 +399,27 @@
...
+
+ @exception `SLUIAElementInvalidException` if element is not valid.
```
wearhere: I should have described the warning better--please use the exception message from `-[SLUIAElement rect]` (what's currently missing is a note that the element doesn't have to be immediately valid).

wearhere:
Thanks for your changes @jzucker2. Two more documentation tweaks and then this looks good. And you're right that this request shouldn't have caused the tests to fail; there's a separate issue (#132) at play. I won't hold up this request on those failures.

jzucker2:
Ok, I think I'm on the same page as you now. Hope this makes more sense. Thanks for the suggestions. Glad to contribute to this project!

wearhere:
:+1: Thanks again!

PR #131, opened 2014-03-03
Modernize the example project.

wearhere:


wearhere:
The failure here was due to #132 and not related to the content of this request.

PR #134, opened 2014-03-08
Pretty-print CI test output.

wearhere:


wearhere:
Even if all the tests pass, this isn't ready to go quite just yet. [The integration test runs](https://travis-ci.org/inkling/Subliminal/jobs/20333012) should use Unicode characters like [the unit test runs](https://travis-ci.org/inkling/Subliminal/jobs/20333008) but don't, so I must not be properly checking for Unicode support.

wearhere:
Ahh, finally. For reference, there remains one formatting discrepancy particular to Travis--test names should be underlined. Filed as travis-ci/travis-ci/issues/2047.

PR #142, opened 2014-03-21
Fix documentation errors

MaxGabriel:
Turning on the the compiler warning for "Documentation Comments" turned up two errors. I ended up leaving that warning flag off because it was triggering errors in the upstream OCMock library, and I wasn't sure how to ignore those. 

edit: Oh, I guess a good fix would be to turn on this compiler warning just for the Subliminal target, rather than for the whole project.

### DIFF
--- a/Sources/Classes/Internal/NSObject+SLVisibility.m
+++ b/Sources/Classes/Internal/NSObject+SLVisibility.m
@@ -51,7 +51,7 @@ const unsigned char kMinVisibleAlphaInt = 3; // 255 * 0.01 = 2.55, but our bitma
  Returns the number of points from a set of test points for which the receiver is visible in a given window.
 
  @param testPointsInWindow a C array of points to test for visibility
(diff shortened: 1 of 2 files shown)


wearhere:
I didn't know that this existed! Neat. This appears to catch errors that Appledoc itself does not catch (when the Subliminal Documentation target is built).

wearhere:
The upstream OCMock library is Inkling's fork, so we could (and probably should) fix the documentation errors there.

However, a bigger issue is that Xcode/Doxygen don't appear to understand variable argument lists--"parameter '...' not found in the function declaration" errors appear everywhere with the setting on. So, I think we shouldn't leave it on, but just toggle it occasionally to double-check Appledoc's results.

Travis says that the documentation target built fine (I really need to figure out why we intermittently see errors in the other builds), so I'm going to merge this.

wearhere:
Thanks!

Issue #144, opened 2014-03-23
Can creating a new info.plist file be removed from the installation process?


MaxGabriel:
It seems like `Integration Tests-Info.plist` file that's added as part of the installation process isn't necessary to use Subliminal (if I'm wrong about this, just ignore this issue). However, a paragraph of instructions and a screenshot are devoted to making sure that it exists. The README goes on to explain how the xcconfig file specifies the path to this plist, and then another screenshot is devoted to deleting a build setting to make the xcconfig file take affect.

Additionally, having two Info.plist files means that you need to keep them in sync, so e.g. when you add a custom font to your main target's plist, you would also need to add it to your Integration Tests target's plist.

Could instead the README recommend that you delete the generated `<target name> copy-Info.plist` file, and that you use the "Choose Info.plist File..." button to select the plist your main target uses? That would solve the syncing problem and would mean you don't have to fiddle with moving the plist file to a different folder and updating Xcode's reference to it.

![screenshot 2014-03-23 11 16 27](https://f.cloud.github.com/assets/1274145/2494102/55d280c6-b2b7-11e3-8783-56729de8b1bd.png)

One downside is that if the xcconfig file were to no longer give the path to the Info.plist, that would be a breaking change for people currently relying on it. One option is to leave this setting in the xcconfig file and to just change the README instructions. The xcconfig file's plist path setting wouldn't take affect if the user selected the plist path manually. 

Alternatively, the xcconfig file could just drop the plist path setting. As far as breaking changes go, its not a big one, as there will be a straightforward compiler error about the problem (additionally, it won't affect people who copied the xcconfig file to their project instead of added it as a reference)

Edit: I guess the same question goes for Product Name? I didn't look into that one.

wearhere:
:+1: (more comments on #145).

Also filed #146 per your suggestion.

Issue #82, opened 2013-09-10
Revise Subliminal's documentation.


wearhere:
- [x] Migrate installation process to wiki page to slim the README

EDIT: This issue used to talk about adding contributing guidelines but that was moved to #166.

MaxGabriel:
Hey @wearhere do you still want to have most of the README moved to the wiki? That'd be pretty easy for me to do.

MaxGabriel:
My plan on moving to the wiki is that I'll move content over to the wiki gradually, where it's available for public review. Unfortunately, I don't think there's any PR system, so I'll just add things and let them be reviewed over time.

Based on the README, I think these are good candidates for being broken out into separate wiki pages:
- [x] [Create the Integration Tests Target page](https://github.com/inkling/Subliminal/wiki/Create-the-Integration-Tests-Target)
- [x] [Install Subliminal with Git submodules](https://github.com/inkling/Subliminal/wiki/Install-Subliminal-With-Git-Submodules-or-Manually)
- [x] [Add Test Running Code to the Target](https://github.com/inkling/Subliminal/wiki/Add-Test-Running-Code-to-the-Target)
  - Added loading Instruments and choosing the Subliminal template
- [x] [Writing Tests](https://github.com/inkling/Subliminal/wiki/Writing-Tests)
  - Added making a new Subliminal test case using the file template
- [x] [Debugging](https://github.com/inkling/Subliminal/wiki/Debugging-Tests)
  - Added a section on logging
- [x] [Continuous Integration](https://travis-ci.org/MaxGabriel/TestTravis)
  - With a section on Travis CI specifically
    - I tested the `.travis.yml` file [with this project](https://travis-ci.org/MaxGabriel/TestTravis)
- [x] Add linkage between wiki pages
- [x] [Wiki front page](https://github.com/inkling/Subliminal/wiki)
  - Missing link to CI page atm
- [x] [Known Issues](https://github.com/inkling/Subliminal/wiki/Known-Issues)
  - I didn't initially plan to add this to the README, but after removing the rest of the README it didn't really fit in. Also I consider it a niche enough topic to not be on the README
- [x] PR to add the images to gh-pages branch; see #161
- [x] PR to remove this stuff from the README; see #160
- [ ] After everything else, update the GH pages branch (a slimmer README would help it alot, so you can actually see the links to the API docs)

[...]

PR #145, opened 2014-03-23
Update README and xcconfig to remove the "Integration Tests-Info.plist" setting

MaxGabriel:
Implementation of #144

The existing README images come from the Subliminal website (which is from the gh-pages branch I assume). I added a directory for documentation images to the master branch, since otherwise I would need a second PR to the gh-pages branch. I can make a separate PR to gh-pages if you want to keep all the images on inkling.github.io/Subliminal

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,18 +103,16 @@ group.
 
 Creating a separate target for your integration tests will allow you to control
 exactly when your tests are run. To begin with, right click your application 
-target and select "Duplicate".
+target and select "Duplicate". Rename your new target to "Integration Tests".
 
 ![](http://inkling.github.io/Subliminal/readme-images/CreateTarget.png)
 
-Next, rename both your new target to "Integration Tests" and the info.plist 
-created for it to "Integration Tests-Info.plist". Move the reference to the 
-newly created target's Info.plist file into your Integration Tests group using the 
-navigator pane, and move the actual Info.plist file into your 
-Integration Tests directory using the Finder. After this step you will need to
-update Xcode's reference to the `plist`.
+Duplicating the target will have generated a `<target name> copy-Info.plist` file.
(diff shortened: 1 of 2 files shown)


MaxGabriel:
Tested this here https://github.com/MaxGabriel/SubliminalSamePlist and Subliminal's working fine.

I did run into several failed attempts at starting the iOS simulator—I think that's just normal simulator wonkiness, though.

wearhere:
I think this is a good idea @MaxGabriel. Anything we can do to slim down the installation instructions is a good idea, and not having to keep two plists in sync is a definite plus. I would like you to add the image to `gh-pages` rather than `master` though, so users don't have to download the documentation images. (The idea behind moving the installation instructions to the Wiki (#82) is that users probably don't need to have them downloaded locally at all, considering that they'll probably install just once, at the same time as they're checking out the repo.)

Also, I'd like to leave the `INFOPLIST_FILE` setting in the `xcconfig` file. Like you suggested, it won't take effect if the user follows your new instructions, so no reason to break existing installations.

Once you've made these changes I'm good to merge. Thanks for your thoroughness in testing as always!

wearhere:
It would be nice to add a comment to the `xcconfig` file about the setting being maintained for historical reasons though.

MaxGabriel:
Awesome. I'll make those changes. I'll probably add in #146 to this PR as well, since some of the README changes can be dropped entirely if #146 is implemented, too.

MaxGabriel:
Ok, should be ready, but depends on #149. This PR also resolves #146

MaxGabriel:
Updated [the test project](https://github.com/MaxGabriel/SubliminalSamePlist) with the new xcconfig file (just copy+pasted it across) and gave it a Product Name so it no longer uses the xcconfig file one. Integration Tests run fine.

wearhere:
Looks good!

PR #150, opened 2014-03-26
Have Travis CI badge use SVG

MaxGabriel:
Looks much better on retina screens.

edit: If you look at the rendered diff it looks a million times better https://github.com/inkling/Subliminal/pull/150/files?short_path=04c6e90#diff-04c6e90faac2675aa89e2176d2eec7d8


wearhere:
Wow yeah. :+1:!

PR #156, opened 2014-03-27
Increment the instruments retry loop counter _after_ it's used.

wearhere:
Previously, we'd skip the first value of the launch timeouts array (30)
and then use a value beyond the array's length, as can be seen in
https://travis-ci.org/inkling/Subliminal/jobs/21707856.


wearhere:
1.1 not because this is a super-critical fix but because it's low-risk and it's been made already. @aegolden / @ahaneyinkling can I get a review?

PR #159, opened 2014-03-29
Remove the text shadowing on paragraphs

MaxGabriel:
The text shadow makes the text look really blurry, especially on the text in the Usage section. Here's a comparison:

[With Shadow](http://inkling.github.io/Subliminal/)
[No Shadow](http://maxgabriel.github.io/Subliminal/)

(I'm on a retina MBP; not sure if that's exaggerating the difference.


wearhere:
Woah I definitely see what you're talking about. Nice catch!

PR #162, opened 2014-03-30
Add the ability to order tests by run group.

wearhere:
See the documentation on `+[SLTest runGroup]`.

This is a proposed resolution to #94 and an alternative to the mechanism implemented in #122.

### DIFF
--- a/Sources/Classes/SLTest.h
+++ b/Sources/Classes/SLTest.h
@@ -80,7 +80,8 @@
  
  An abstract test will not itself be run. Subclasses which do define test cases
  will be run, however, allowing a single base class to define set-up and tear-down 
- work shared among related subclasses.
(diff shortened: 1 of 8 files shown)


wearhere:
[Ignoring whitespace](https://github.com/inkling/Subliminal/pull/162/files?w=1) helps when looking at the changes to `SLTestController.m`.

j-mutter:
I like it. :+1:

My solution was a total hack to just get the tests running in a way we needed, and as you pointed out caused a few quirks around the startup and teardown of the test suites, etc.

wearhere:
I'm glad to hear you like it @j-mutter! Let's get this into 1.1 then; we'll finally have a solution for people's startup woes. Can I get a review @aegolden?

Issue #164, opened 2014-03-30
dispatch_get_current_queue is deprecated in iOS 6


MaxGabriel:
If you set the deployment target of the Subliminal Project to 6.0+ and then try to build the Subliminal library, there are many deprecation notices for `dispatch_get_current_queue` being deprecated. Based on a quick look, it looks like Subliminal mostly uses that function for the pattern 'if I am not on that queue, dispatch sync to that queue, otherwise just execute normally'. 

I think the proposed workaround for that pattern is to use the `dispatch_queue_set_specific` and `dispatch_get_specific` APIs:

``` objc
@interface ViewController ()
@property (nonatomic, strong) dispatch_queue_t myQueue;
@end

@implementation ViewController

BOOL isMyQueue(void) {
    return dispatch_get_specific(@selector(myQueue)) != NULL;
}

- (void)viewDidLoad
{
    [super viewDidLoad];
    NSLog(@"Loading");


    self.myQueue = dispatch_queue_create("myqueue", DISPATCH_QUEUE_SERIAL);
    dispatch_queue_set_specific(self.myQueue, @selector(myQueue), @selector(myQueue), NULL);
    NSLog(@"Is My Queue from main = %i",isMyQueue()); // False

    dispatch_async(self.myQueue, ^{
        NSLog(@"Is My Queue from async = %i",isMyQueue()); // True
    });
}
```

References:

http://stackoverflow.com/a/12806754/1176156
https://devforums.apple.com/thread/162529?start=0&tstart=0

(My implementation is mostly cropped from the second page of the dev forums post, except I used the trick of [using a selector for the `const void *`](http://nshipster.com/associated-objects/))

wearhere:
Nice find! I had previously seen the warning, done some searches, and found only [answers](http://stackoverflow.com/questions/13237417/alternatives-to-dispatch-get-current-queue-for-completion-blocks-in-ios-6) where people were like "why were you using `dispatch_get_current_queue` to begin with." >.< (See also the first page of that dev forum discussion, sheesh.)

It'd be good to pick this up sometime in the near future, certainly before iOS 8. I endorse your approach except for that I prefer

``` objc
static const void *const kIsMyQueueKey = &kIsMyQueueKey;
```

to using selectors as keys. While convenient, selectors don't guarantee that you use the same key across API calls--you could easily mistype a selector, especially if you have the "unknown selector" warning turned off (as Subliminal does, so that the compiler doesn't complain about the use of app hooks).

[...]

PR #148, opened 2014-03-25
Abort and retry when `instruments` launches the simulator in an inconsistent state.

wearhere:
Works around #147.

Also fixes a bug where we didn't maintain the previously set device type/version when we reset the simulator.

Issue #71, opened 2013-08-21
Update Subliminal to use appledoc 2.2.


wearhere:
At the least, this will entail that we document enum values and bitmasks (tomaz/appledoc#378).

wearhere:
We should wait to tackle this until there's an explanation of how to document enum values and bitmasks beyond [this screenshot](https://github.com/tomaz/appledoc/issues/2#issuecomment-21579308). [Appledoc's documentation](http://gentlebytes.com/appledoc/) is currently down…

It would also be nice to update `.travis.yml` to use an archived release of Appledoc 2.2 (e.g. Homebrew/homebrew-versions#204) rather than have to point to a specific (new) version of the main Appledoc formula like we're doing [currently](https://github.com/inkling/Subliminal/blob/master/.travis.yml#L11).

wearhere:
This is worth doing for 2.0, but only so that we're "completely" documented before switching to Cocoadocs (#195)--and if this turns out to be a huge pain (considering that [Appledoc's documentation](http://gentlebytes.com/appledoc/) _still_ isn't back up), we should let this go.

PR #168, opened 2014-03-31
Use dispatch_get_specific to workaround the deprecation of dispatch_get_current_queue

MaxGabriel:
Addresses #164

### DIFF
--- a/Logging/SLLogger.h
+++ b/Logging/SLLogger.h
@@ -111,6 +111,15 @@ void SLLogAsync(NSString *format, ...) NS_FORMAT_FUNCTION(1, 2);
 - (dispatch_queue_t)loggingQueue;
 
 /**
+ Whether or not the current queue is the `loggingQueue`.
+ 
+ To avoid deadlocks, use this method to check if you're on the `loggingQueue` before `dispatch_sync`ing a block to it.
+ 
+ @return Whether or not the current queue is the `loggingQueue`.
+ */
+- (BOOL)currentQueueIsLoggingQueue;
+
+/**
  Logs a message.
  
  This method is the primitive logging method used by all other logging methods 
--- a/Sources/Classes/Internal/Terminal/SLTerminal.h
+++ b/Sources/Classes/Internal/Terminal/SLTerminal.h
@@ -140,6 +140,15 @@
 @property (nonatomic, readonly) dispatch_queue_t evalQueue;
 
 /**
+ Whether or not the current queue is the `evalQueue`.
+ 
+ To avoid deadlocks, use this method to check if you're on the `evalQueue` before `dispatch_sync`ing a block to it.
+ 
+ @return Whether or not the current queue is the `evalQueue`.
+ */
+- (BOOL)currentQueueIsEvalQueue;
+
+/**
  Causes `SLTerminal.js` to finish evaluating commands.
 
  The terminal starts up automatically when the UIAutomation instrument is attached
--- a/Sources/Classes/Internal/Terminal/SLTerminal.m
+++ b/Sources/Classes/Internal/Terminal/SLTerminal.m
@@ -48,6 +48,11 @@ const NSTimeInterval SLTerminalReadRetryDelay = 0.1;
 // is closer to 0.05.
 const NSTimeInterval SLTerminalEvaluationDelay = 0.075;
 
+/**
+ Identifier for the `evalQueue` for use with `dispatch_get_specific`.
+ */
+static const void *const kEvalQueueIdentifier = &kEvalQueueIdentifier;
+
 @implementation SLTerminal {
     NSString *_scriptNamespace;
     dispatch_queue_t _evalQueue;
@@ -81,6 +86,7 @@ static SLTerminal *__sharedTerminal = nil;
     if (self) {
         _scriptNamespace = SLTerminalNamespace;
         _evalQueue = dispatch_queue_create("com.inkling.subliminal.SLTerminal.evalQueue", DISPATCH_QUEUE_SERIAL);
+        dispatch_queue_set_specific(_evalQueue, kEvalQueueIdentifier, (void *)kEvalQueueIdentifier, NULL);
     }
     return self;
 }
@@ -101,6 +107,11 @@ static SLTerminal *__sharedTerminal = nil;
 
 - (dispatch_queue_t)evalQueue {
     return _evalQueue;
+}
+
+- (BOOL)currentQueueIsEvalQueue
+{
+    return dispatch_get_specific(kEvalQueueIdentifier) != NULL;
 }
 
 #if TARGET_IPHONE_SIMULATOR
@@ -161,7 +172,7 @@ static SLTerminal *__sharedTerminal = nil;
     NSParameterAssert(script);
     NSAssert(![NSThread isMainThread], @"-eval: must not be called from the main thread.");
 
-    if (dispatch_get_current_queue() != self.evalQueue) {
+    if (![self currentQueueIsEvalQueue]) {
         id __block result;
         NSException *__block evalException;
         dispatch_sync(self.evalQueue, ^{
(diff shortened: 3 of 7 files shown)


Sources/Classes/Internal/Terminal/SLTerminal.m
```diff
@@ -81,6 +86,7 @@ - (id)init {
     if (self) {
         _scriptNamespace = SLTerminalNamespace;
         _evalQueue = dispatch_queue_create("com.inkling.subliminal.SLTerminal.evalQueue", DISPATCH_QUEUE_SERIAL);
+        dispatch_queue_set_specific(_evalQueue, kEvalQueueIdentifier, (void *)kEvalQueueIdentifier, NULL);
```
MaxGabriel, line 89: If you don't cast to `(void *)` then you get a warning that it discards the `const` qualifier. I'm not sure if just silencing the warning via cast is the best move.

(This also applies to `SLLogger`)

Sources/Classes/Internal/Terminal/SLTerminal.m
```diff
@@ -103,6 +109,11 @@ - (dispatch_queue_t)evalQueue {
     return _evalQueue;
 }
 
+- (BOOL)currentQueueIsEvalQueue
+{
+    return dispatch_get_specific(kEvalQueueIdentifier) != NULL;
+}
```
MaxGabriel, line 115: I made these methods instance methods since they depend on the `[SLTerminal sharedInstance]` being initialized, since that's what initializes the queue.

Sources/Classes/Internal/Terminal/SLTerminal.h
```diff
@@ -140,6 +140,13 @@
 @property (nonatomic, readonly) dispatch_queue_t evalQueue;
 
 /**
+ *  Whether or not the current queue is the `evalQueue`. This method works around the deprecation of `dispatch_get_current_queue`.
+ *
+ *  @return Whether or not the current queue is the `evalQueue`.
+ */
+- (BOOL)currentQueueIsEvalQueue;
+
+/**
```
MaxGabriel, line 151: Not sure whether this should be private API. Its somewhat convenient to be public b/c it makes it easier for `SLTerminal+ConvenienceFunctions` to use. 

Its offering no more functionality than was previously offered by `dispatch_get_current_queue != [SLTerminal sharedTerminal].evalQueue` so its definitely safe, its just a question of whether it makes sense to be public.

Logging/SLLogger.h
```diff
@@ -111,6 +111,13 @@
 - (dispatch_queue_t)loggingQueue;
 
 /**
+ *  Whether or not the current queue is the `loggingQueue`. This method works around the deprecation of `dispatch_get_current_queue`.
+ *
+ *  @return Whether or not the current queue is the `loggingQueue`.
+ */
+- (BOOL)currentQueueIsLoggingQueue;
```
MaxGabriel, line 120: I don't quite like this name, but I haven't come up with anything better.

wearhere, line 120: Agree but agree. :)

Sources/Classes/Internal/Terminal/SLTerminal.h
wearhere, line 151: It's in a category marked/documented as "Internal" (alongside `evalQueue`, as you noted) so I don't feel terrible about it being here. Now, I'm not sure why I declared the "Internal" category in the main header, heh--I think because `SLTerminal` itself is kinda internal to Subliminal (the header is public but requires specific import--it's not included in `Subliminal.h`), and I figured that anyone who's using it could be trusted to read the header.

Logging/SLLogger.h
```diff
@@ -111,6 +111,13 @@
 - (dispatch_queue_t)loggingQueue;
 
 /**
+ *  Whether or not the current queue is the `loggingQueue`. This method works around the deprecation of `dispatch_get_current_queue`.
```
wearhere: Let's not mention the deprecation here/on `-[SLTerminal currentQueueIsEvalQueue]`--someone can check the commit messages if they want to see why this was added, and at some point `dispatch_get_current_queue` won't even exist anymore.

I do think it would make sense to say why this these methods are here at all though, i.e. to write re-entrant APIs using the queues.

Sources/Classes/Internal/Terminal/SLTerminal.m
wearhere, line 89: I think casting is alright. I'm not sure what else we'd use as the value, and the value doesn't really matter anyway.

Sources/Classes/Internal/Terminal/SLTerminal.m
wearhere, line 115: Makes sense.

Unit Tests/SLDispatchTests.m
```diff
@@ -0,0 +1,44 @@
...
+#import "SLTerminal.h"
+#import "SLLogger.h"
+
+/**
+ *  When Apple deprecated the `dispatch_get_current_queue` functionality, the workaround we came up with was to use `dispatch_queue_set_specific` and `dispatch_get_specific` to check if we were on that queue. 
+ 
+ These tests verify that was a valid workaround.
+ 
+ See https://github.com/inkling/Subliminal/issues/164
+ */
```
wearhere, line 19: Adding tests was a nice gesture, thanks. (And since the tests are intimately related to the development of the API I think mentioning `dispatch_get_current_queue` here makes sense, fwiw).

Logging/SLLogger.h
MaxGabriel: Ok will do.

wearhere:
Thanks as always for writing docs @MaxGabriel. Can you match the prevailing style in the project not to begin lines with asterisks though please?

``` objc
/**
 This is indented alongside the beginning/ending asterisks but does not itself begin with an asterisk.
 */
```

EDIT: Changed second sentence to say please _do_ match the prevailing style, haha.

Logging/SLLogger.h
MaxGabriel: Does this sound right?

_Whether or not the current queue is the `loggingQueue`. Use this method to check if you're on the `loggingQueue` before `dispatch_sync`ing a block to it to avoid deadlocks._

MaxGabriel:
Yeah @wearhere I was just using [VVDocumenter](https://github.com/onevcat/VVDocumenter-Xcode) and that's how they do the autocompletion for you

Logging/SLLogger.h
wearhere: I personally would write "To avoid deadlocks, use this method..." and also put a blank line between the first and second sentence (so Appledoc puts the second sentence in the method's "Discussion" section), but essentially yes. :)

wearhere:
Oh that's a neat project, I hadn't seen that before. I especially like that it'll generate `NS_ENUM` docs since Appledoc's documentation is _still_ [down](http://gentlebytes.com/appledoc/) and the lack of enum documentation has held up our adoption of the latest version of Appledoc (#71).

MaxGabriel:
@wearhere Ok, implemented the suggestions

wearhere:
Looks good @MaxGabriel. One more thing--can you squash the third commit into the second? The latter (mediation API?) doesn't seem to be related to this request.

MaxGabriel:
Yeah woops, pulled up the wrong terminal when I was running git commit at work.

MaxGabriel:
@wearhere Have you seen Travis stall out before on:

```
SLKeyboardTest started.
        │ "testTypeStringChangesKeyplanesAsNecessary" started.
```

Not sure if it was related to my changes.

wearhere:
@MaxGabriel Rats, that looks like the sort of error that #148 was intended to fix--guess it wasn't completely sufficient (though I think the builds have got a lot better).

In any case, I doubt it's related to your changes. I respun the job and I bet this one will go just fine.

MaxGabriel:
Thanks @wearhere

wearhere:
Passed! Thanks @MaxGabriel, merging.

PR #170, opened 2014-04-01
Cast to NSURL to silence warning when using Cocoapods

### LINKED ISSUE #169: Erroneous incompatible pointer types warning when installing w/ Cocoapods
When installing with Cocoapods I get this warning in two places (`SLTestController` and `SLTerminal`):

SLTestController.m:254:15: 

Incompatible pointer types initializing 'NSString *__strong' with an expression of type 'CGPathRef' (aka 'const struct CGPath *')

which is caused by the following code:

``` objc
NSString *userDirectoryPath = [[[[NSFileManager defaultManager] URLsForDirectory:NSLibraryDirectory inDomains:NSUserDomainMask] lastObject] path];
```

(The compiler is thinking that the `path` selector is the one from `CAKeyFrameAnimation`, based on the documentation it pulls up when option clicking).

Fixing this warning is super trivial—I just add a cast to `(NSURL *)`. Is that solution OK? I've tried to figure out what the root cause of why this warning shows up when installing with Cocoapods but not normally, but I haven't come up with anything.

MaxGabriel:
Closes #169

### DIFF
--- a/Sources/Classes/Internal/Terminal/SLTerminal.m
+++ b/Sources/Classes/Internal/Terminal/SLTerminal.m
@@ -126,7 +126,7 @@ static SLTerminal *__sharedTerminal = nil;
         NSString *plistName = [NSString stringWithFormat:@"%@.plist", [[NSBundle mainBundle] bundleIdentifier]];
 
         // 1. get into the simulator's app support directory by fetching the sandboxed Library's path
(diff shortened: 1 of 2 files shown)


wearhere:
Thanks for your investigation and fix @MaxGabriel!

PR #171, opened 2014-04-03
Accessibility parent

jzucker2:
This is mostly a rough draft, wanted to see what you thought of this API for matching child elements.


jzucker2:
I know it's missing documentation, if you like the approach, I'll add docs and resubmit. If not, we can figure out something a little different.

wearhere:
Thanks for the PR @jzucker2. Perfectly reasonable to submit a WIP. What would be helpful instead of docs though would be a little bit (here in comments) about the problem that the new API's solving. We've talked about it before but formalizing it here will help me understand the request and best comment on its direction.

jzucker2:
This is a formal addition to Subliminal for matching and testing against elements inside a UITableViewCell. Subliminal does not have great abstraction for child and parent elements. This formalizes that in an easy API.

My big concern is how I fixed the slAccessibilityIsVisible method, by comparing against UITableTextAccessibilityElement, which isn't very clean, but it was an awkward comparison there. Really looking for feedback on that.

As well, this can be extended with a subclass for UICollectionViewCell, but I waited to do that until I heard about the UITableViewCell patch.

And technically, this API could be extended to any elements inside another view or accessibility container.

jzucker2:
This is a cleaner implementation. The creation of SLAccessibilityContainer is a little awkward, but I had to get around initWithPredicate being a private method. This works though, except it struggles with isVisible because it seems to only select the mock accessibility view and not the actual view for isVisible. I fixed with a small hack but would love a stronger solution.

Sources/Classes/Internal/NSObject+SLVisibility.m
```diff
@@ -133,7 +133,7 @@ - (BOOL)slAccessibilityIsVisible {
             // if another element comes before us/our parent in the array
             // (thus is z-ordered before us/our parent)
             // and contains our hitpoint, it covers us
-            if (CGRectContainsPoint([element accessibilityFrame], testPoint)) return NO;
+            if (![NSStringFromClass([element class]) isEqualToString:@"UITableTextAccessibilityElement"] && CGRectContainsPoint([element accessibilityFrame], testPoint)) return NO;
```
jzucker2, line 136: This is the one issue I haven't been able to figure out for myself.

jzucker2:
The failure is from lack of documentation, not any compilation/test failures.

jzucker2:
I'm closing this pull request in favor of a cleaner implementation that I just submitted: https://github.com/inkling/Subliminal/pull/190

PR #173, opened 2014-04-06
Add delayed show method for UIAlertView.

TadeasKriz:
I've added a category for `UIAlertView` that will delay the `-[UIAlertView show]` method until the default alert handling is set up. This way if a dialog is to be shown right after the application is started, it will delay it until Subliminal is loaded.

### DIFF
--- a/Sources/Classes/Internal/UIAlertView+SLDelayShow.h
+++ b/Sources/Classes/Internal/UIAlertView+SLDelayShow.h
@@ -1,0 +1,10 @@
+//
+// Created by Tadeas Kriz on 06/04/14.
+// Copyright (c) 2014 Inkling. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+#import <UIKit/UIKit.h>
+
+@interface UIAlertView (SLDelayShow)
+@end
--- a/Sources/Classes/Internal/UIAlertView+SLDelayShow.m
+++ b/Sources/Classes/Internal/UIAlertView+SLDelayShow.m
@@ -1,0 +1,53 @@
+//
+// Created by Tadeas Kriz on 06/04/14.
+// Copyright (c) 2014 Inkling. All rights reserved.
+//
+
+#import <objc/runtime.h>
+#import "UIAlertView+SLDelayShow.h"
+#import "SLAlert.h"
+#import "SLLogger.h"
+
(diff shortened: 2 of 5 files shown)


wearhere:
Hi @TadeasKriz, thanks for the PR. You're right that it's a problem for alerts to show immediately on startup, and this is a neat idea to work around that.

Unfortunately swizzling `-show` doesn't appear to catch one of the alerts most likely to appear on launch, the location permissions alert. Check out https://github.com/inkling/Subliminal/tree/test/location_alert_on_startup.

I also note that this solution only handles alerts being automatically dismissed--`+[SLAlertHandler addHandler:]` can't be called before the alert would be shown (from the main thread), and even if it could, it would be tricky to later verify that the alert was handled (in the context of a test).

I wonder if it might just be better to encourage people to wait to show alerts until testing has actually started? That's our [current recommendation](http://stackoverflow.com/questions/18479087/handle-location-permission-popup-with-subliminal).

TadeasKriz:
I've read that recommendation, but the problem is that I don't show the alert in `-application:didFinishLaunchingWithOptions:`, but in '-viewDidLoad' of my initial view controller. 

So I was thinking (but didn't get to investigate it properly, if it's even possible, for example I'm not really sure, when is initial ViewController presented when using storyboards, if it's before or after the `-application:didFinishLaunchingWithOptions:`), that calling "runTests" might stall the code execution until instrumentations are loaded. What do you think?

I'd probably close this PR, because even to me it's of no value now (I've changed my tests to set clean rootViewController for each test method, so the dialog in viewDidLoad is of no problem anymore).

wearhere:
The recommendation describes any alerts that might be shown before Subliminal starts up, not just alerts shown in `-application:didFinishLaunchingWithOptions:`.

I haven't used storyboards but I would bet that the initial view controller is presented after `-application:didFinishLaunchingWithOptions:`. `-runTests...` doesn't block (the tests are scheduled on a background thread) so there's no guarantee that Subliminal would have loaded its alert handling before `-viewDidLoad` was executed. The best way to guarantee that the alert could be handled would definitely be to show it in response to some test action--either simulated user interaction, or the use of an app hook.

Anyway I'll close this PR like you suggest--glad to hear you're no longer having problems. Let me know if you have any further questions though.

PR #179, opened 2014-04-11
Include collection view cell mock views in accessibility paths.

### LINKED ISSUE #178: UICollectionViewCell filtered out of Accessibility Element Path
I ran into an issue with trying to click a button that is a child of a UICollectionViewCell,  when the SLAccessibilityPath builds up it's path filterRawAccessibilityElementPath skips over the cell in the hierarchy which causes the button to never become tappable.

I fixed it with a quick hack by adding:

if ([self isKindOfClass:[UICollectionViewCell class]]) {
        return YES;
}

in NSObject+SLAccessibilityHierarchy - (BOOL)classForcesPresenceInAccessibilityHierarchy

wearhere:
Fixes #178.

### DIFF
--- a/Integration Tests/Tests/SLElementMatchingTest.m
+++ b/Integration Tests/Tests/SLElementMatchingTest.m
@@ -34,6 +34,16 @@
     return @"SLElementMatchingTestViewController";
 }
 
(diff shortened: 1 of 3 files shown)


wearhere:
Adding this to 1.1, and removing #178, just to clarify when looking at PRs.

Issue #135, opened 2014-03-09
Subliminal incorrectly thinks elements are invisible in non-portrait orientations


MaxGabriel:
Hi, I think there's a bug where Subliminal incorrectly labels an element invisible in non-portrait orientations. I created [a sample project](https://github.com/MaxGabriel/SubliminalLandscapeBug) replicating this bug. 

I believe the problem comes from [this block of code](https://github.com/inkling/Subliminal/blob/master/Sources/Classes/Internal/NSObject%2BSLVisibility.m#L300-L316). In landscape, Subliminal considers the center to be covered, but in portrait it thinks it is not.

I tried to investigate down into `numberOfPointsFromSet:...` and `renderViewRecursively:...`, but got lost in the Core Graphics code. Any idea what the problem could be? My suspicion is that its a screen coordinates != logical coordinates type of issue, but I'm not sure.
#### Steps to replicate
1. `git clone https://github.com/MaxGabriel/SubliminalLandscapeBug.git`
2. `cd SubliminalLandscapeBug`
3. `git submodule update --init`
4. `open Subliminal.xcodeproj`
5. Run integration tests target (I ran against the iOS 7 4" simulator)

EDIT (@wearhere): This issue originally described the failure as occurring in landscape orientations, but tests added in #180 revealed that the failure actually occurs in any non-portrait orientation. This issue's title and description have been updated accordingly.

MaxGabriel:
Doubt it'd be helpful, but here's the trace file from the Instruments run https://s3.amazonaws.com/uploads.hipchat.com/1658/151914/AnCa4c6gojjpUMA/SubliminalLandscapeTrace.trace.zip

wearhere:
Thanks for the terrific bug report @MaxGabriel! I can reproduce. I imagine that it’s a matter of needing to transform the context into which the view is rendered, but I'm not quite sure how to do that myself... @aegolden can you take a look?

wearhere:
Hey @MaxGabriel, we're looking into the bug. In the meantime, you can fall back on UIAutomation's visibility routine using the following:

``` objc
#import <Subliminal/SLUIAElement+Subclassing.h>

@interface SLUIAElement (LandscapeVisibility)
- (BOOL)uiaIsVisible;
@end

@implementation SLUIAElement (LandscapeVisibility)
- (BOOL)uiaIsVisible {
    return [[self waitUntilTappable:NO thenSendMessage:@"isVisible()"] boolValue];
}
@end
```

[...]

PR #180, opened 2014-04-12
Workaround the non-portrait-orientation bug in our visibility routine.

wearhere:
While we work on #135, the workaround described on that bug should make it into 1.1. cc @MaxGabriel

### DIFF
--- a/Integration Tests/Tests/SLDeviceTest.m
+++ b/Integration Tests/Tests/SLDeviceTest.m
@@ -90,6 +90,14 @@
     }
 }
 
+// Exercises a bug in iOS 5.1 and 7.1. See https://github.com/inkling/Subliminal/pull/180#issuecomment-40891098
+- (void)testThatDeviceOrientationPersistsThroughDeactivation {
+    UIDeviceOrientation currentOrientation = [UIDevice currentDevice].orientation;
+    [[SLDevice currentDevice] deactivateAppForDuration:3.0];
+    SLAssertTrue([UIDevice currentDevice].orientation == currentOrientation,
+                 @"The device orientation should not have changed.");
+}
+
 - (void)rotateToAndCheckOrientation:(UIDeviceOrientation)orientation
 {
     [[SLDevice currentDevice] setOrientation:orientation];
--- a/Integration Tests/Tests/SLElementVisibilityTest.m
+++ b/Integration Tests/Tests/SLElementVisibilityTest.m
@@ -80,7 +80,23 @@
 
     if (testSelector == @selector(testCanDetermineVisibilityOfWebAccessibilityElements)) {
         SLAssertTrueWithTimeout(SLAskAppYesNo(webViewDidFinishLoad), 5.0, @"Webview did not load test HTML.");
+    } else if ((testSelector == @selector(testViewIsVisibleInPortrait)) ||
+               (testSelector == @selector(testViewIsVisibleInPortraitUpsideDown)) ||
+               (testSelector == @selector(testViewIsVisibleInLandscapeLeft)) ||
+               (testSelector == @selector(testViewIsVisibleInLandscapeRight))) {
+        SLAskApp(showTestView);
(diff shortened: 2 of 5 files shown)


MaxGabriel:
Cool I can confirm this fixes the failing tests in the [sample project](https://github.com/MaxGabriel/SubliminalLandscapeBug) that reproed the bug.

MaxGabriel:
For the `setOrientation` timer thing, do you maybe want something that periodically checks the orientation returned from `[[UIDevice currentDevice] orientation]`? That way you could use a longer timeout and raise a clear exception if rotation failed. (Not sure why it'd be this unreliable though)

``` objc
waitUntil(^BOOL{
        return [[UIDevice sharedDevice] orientation] == orientation;
    }, 4);
```

``` objc
void waitUntil(BOOL (^waitBlock)(void), const NSTimeInterval timeout)
{
    NSCParameterAssert(waitBlock);
    NSCParameterAssert(timeout > 0);

    NSTimeInterval timeWaited = 0;
    while (true) {

        __block BOOL waitCondition = NO;
        dispatchSyncMainQueueIfNecessary(^{
            waitCondition = waitBlock();
        });

        if (waitCondition) {
            break;
        } else if (timeWaited >= timeout) {
            NSCAssert(@"Timeout exception", @"Hit timeout waiting for an event");
        } else {
            static const NSTimeInterval sleepInterval = 0.3;
            [NSThread sleepForTimeInterval:sleepInterval];
            timeWaited += sleepInterval;
        }
    }
}

void dispatchSyncMainQueueIfNecessary(void (^block)(void))
{
    NSCParameterAssert(block);
    if ([NSThread isMainThread]) {
        block();
    } else {
        dispatch_sync(dispatch_get_main_queue(), ^{
            block();
        });
    }
}
```

wearhere:
Oh, I think that the orientation does reliably update to that set--within the old time interval, actually, as the rotation test passes just fine. I just want to make sure that the rotation _animation_ completes by the time that `-setOrientation:` returns, 'cause otherwise it'll look a bit weird from a user's standpoint and there might be problems with visibility/tappability.

I'm not sure why the tests failed the last time, but I don't think it was due to the orientation not being set right.

wearhere:
Found out why they failed before--on iOS 5.1 and 7.1, iPads forget their orientation when deactivated! The orientation actually becomes `UIDeviceOrientationUnknown`.

That's why [this test run](https://travis-ci.org/inkling/Subliminal/jobs/22828382) failed--when `-[SLDeviceTest testDeactivateAppForDuration]` ran before the visibility test, all the visibility test cases that test differences in behavior between Subliminal and UIAutomation started failing, because `-[SLElement isVisible]` thought that the device was in a non-portrait orientation and so used `UIAElement.isVisible` rather than Subliminal's visibility routine.

EDIT: Since this comment is linked from the commit history, I've clarified that it is _iPads_ running iOS 5.1 _or_ 7.1 that exhibit this issue.

wearhere:
Aaand `UIDevice` forgets on iOS 7.1 too >.<.

MaxGabriel:
I wasn't able to replicate it, but maybe calling `beginGeneratingDeviceOrientationNotifications` before this test would fix it?

wearhere:
Further clarification: I see `-[SLDeviceTest testThatDeviceOrientationPersistsThroughDeactivation]` consistently fail on the _iPad_ when running either iOS 5.1 or 7.1 (but 6.1's fine).

I don't doubt that it might work without the reset in some circumstances--in [this run](https://travis-ci.org/inkling/Subliminal/jobs/23372969), _some_ of the visibility test cases failed in a way that would suggest that the orientation had been forgotten, but not all, and `-[SLDeviceTest testThatDeviceOrientationPersistsThroughDeactivation]` passed there.

Calling `-[UIDevice beginGeneratingDeviceOrientationNotifications]` didn't seem to help fwiw.

PR #191, opened 2014-04-27
Add SLSlider

MaxGabriel:
edit: ignore for now; there appears to be some weird iOS 7 UIASlider behavior going on...

Adds `SLSlider` to the family of `SLElement` subclasses

### DIFF
--- a/Sources/Classes/UIAutomation/User Interface Elements/SLSlider.h
+++ b/Sources/Classes/UIAutomation/User Interface Elements/SLSlider.h
@@ -1,0 +1,47 @@
+//
+//  SLSlider.h
+//  Subliminal
+//
+//  Created by Maximilian Tagher on 4/27/14.
+//  Copyright (c) 2014 Inkling. All rights reserved.
+//
+
+#import <Subliminal/Subliminal.h>
+#import "SLUIAElement+Subclassing.h"
+
+/**
+ `SLSlider` matches against instances of `UISlider`.
+ 
+ The slider's thumb can be dragged to a specified value, and the value can be queried using `floatValue` or `value`.
+ */
+@interface SLSlider : SLElement
+
+/**
+  Drags the slider to the specified value.
+ 
+ @param value The desired decimal value from 0 to 1, inclusive. A 0 value represents far left and a value of 1 represents far right.
+ 
+ @exception NSInternalInconsistencyException if `value` is not between 0 and 1, inclusive.
+ */
+- (void)dragToValue:(float)value;
+
+/**
+ The current value of the slider.
+ 
+ @return A float between 0 and 1, inclusive.
+ */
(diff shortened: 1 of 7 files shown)


Sources/Classes/UIAutomation/User Interface Elements/SLSlider.h
```diff
@@ -0,0 +1,47 @@
...
+ @exception NSInternalInconsistencyException if `value` is not between 0 and 1, inclusive.
+ */
+- (void)dragToValue:(float)value;
+
+/**
+ The current value of the slider.
+ 
+ @return A float between 0 and 1, inclusive.
+ */
+- (float)floatValue;
```
MaxGabriel, line 33: So I found it surprising that while `dragToValue:` and various `UISlider` methods take a float between 0 and 1, `value` returns a string like "100%". I added `floatValue` to allow the slider's value to be queried in a way consistent with `dragToValue:`.

The naming is still pretty confusing, though. Is there a better way?

I could, for example, change `dragToValue` to take a value between 0 and 100 and divide that value before passing it off to UIAutomation, but this differs from the `UISlider` behavior of taking a value between 0 and 1. 

I could also override `value` to return a string between 0 and 1, with no percent sign. It doesn't look like any of the other `SLElement` subclasses change the value that UIAutomation returns, though.

Sources/Classes/UIAutomation/User Interface Elements/SLSlider.h
```diff
@@ -0,0 +1,47 @@
...
+ Returns the slider's value as a percentage between 0 and 100, inclusive.
+ 
+ @warning The returned value is between 0 and 100, whereas `dragToValue` is between 0 and 1. Use `floatValue` to get a value between 0 and 1.
+ 
+ @return A string between 0 and 100 with a percent sign, e.g. @"100%".
+ 
+ @exception SLUIAElementInvalidException Raised if the element is not valid
+ by the end of the [default timeout](+defaultTimeout).
+ */
+- (NSString *)value;
```
MaxGabriel, line 45: Given what I considered to be an unexpected return value, I overrode this method to add subclass-specific documentation (the implementation just calls `super`).

MaxGabriel:
I ran into a bug with `UIASlider`'s `dragToValue` function on iOS 7. I reported it as rdar://16738650. It is also available on [Open Radar](http://openradar.appspot.com/radar?id=5816625622155264).

MaxGabriel:
If someone should need this functionality, its possible to somewhat workaround the the bug by using `dragWithStartOffset:endOffset:`. However, this is not as precise as the correctly working UIAutomation `dragToValue` function (e.g. when dragging to 0.0, the slider's `value` method returns 1%. When dragging to 0.2, the slider's `value` method returns 21%).

PR #200, opened 2014-05-09
Add ability to use FOCUS environment variable to focus tests

qiffp:
We found that our CI test runs were getting really long and wanted to split them up into multiple jobs, so we came up with this and have been using it for 3 months without any issues. It has the same behaviour as prepending tests with `focus_`, but it works using a `FOCUS` environment variable on the Integration Tests scheme, which can then be set for each CI job with `subliminal-test [...] -e FOCUS testFoo,testBar`.

We couldn't figure out a way to write tests for this, given that it's checking the environment variables for the list of tests to run, but let me know if you can think of something.

### DIFF
--- a/Sources/Classes/SLTest.h
+++ b/Sources/Classes/SLTest.h
@@ -151,6 +151,43 @@
  */
 + (BOOL)isFocused;
 
+/**
+ Returns YES if the test has at least one test case which is focused
+ using an environment variable named `FOCUS` on the tested application's Integration
+ Tests scheme and which can run on the current platform.
+
+ When a test is run, if any of its test cases are focused, only those test cases will run.
+ This may be useful when writing or debugging tests.
+
+ In Xcode, test cases are focused by listing the desired tests, separated
+ by commas, as the Integration Tests scheme's `FOCUS` environment variable's value.
+
+ Using the subliminal-test script to run Subliminal from the command line for continuous
+ integration, test cases are focused like so:
+
+ subliminal-test [...] -e FOCUS testFoo,testBar
+
+ It is also possible to implicitly focus all test cases by listing their test's name
+ in the `FOCUS` environment variable. But if some test cases are explicitly focused
+ (as above), only those test cases will run--the narrowest focus applies.
+
+ If a test is focused, that focus will apply to any tests which descend from it.
+
+ @warning Methods that take test case selectors as arguments (like
+ `-setUpTestCaseWithSelector:`) are invoked with the unfocused form of the selectors
+ --they need not (and should not) be modified when a test case is focused.
+
+ @warning Focused test cases will not be run if their test is not run (e.g. if
+ it is not included in the set of tests to be run, or if it does not support
+ the current platform).
+
+ @return `YES` if any test cases are focused and can be run on the current platform,
+ `NO` otherwise.
+
+ @see -[SLTestController runTests:usingSeed:withCompletionBlock:]
+ */
++ (BOOL)isFocusedWithEnvVar;
+
 #pragma mark - Ordering Test Runs
 /// ------------------------------------------
 /// @name Ordering Test Runs
--- a/Sources/Classes/SLTest.m
+++ b/Sources/Classes/SLTest.m
@@ -95,6 +95,16 @@ const NSTimeInterval SLWaitUntilTrueRetryDelay = 0.25;
     return NO;
 }
 
++ (BOOL)isFocusedWithEnvVar {
+    for (NSString *testCaseName in [self envVarFocusedTestCases]) {
+        SEL testCaseSelector = NSSelectorFromString(testCaseName);
+        if ([self testCaseWithSelectorSupportsCurrentPlatform:testCaseSelector]) {
+            return YES;
+        }
+    }
+    return NO;
+}
+
 + (BOOL)supportsCurrentPlatform {
     // examine whether this test or any of its superclasses are annotated
     // the "nearest" annotation determines support ("nearest" like with method overrides)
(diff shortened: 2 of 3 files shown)


MaxGabriel:
So I don't really know much about your setup, but is this scalable for you guys? It seems like if you're adding a new test case you have to update either some CI config file (which people might forget to do) or the CI machine has to parse out all the test names or something.

Personally I think this is a little specific to your use case, in light of this being mostly accomplishable by filtering the tests passed to `runTests:withCompletionBlock:`, and there being multiple approaches you might take to choose which tests to run (see below).

On the other hand, none of these approaches are mutually exclusive, and I know Jeff says multiple people are now doing something like this, so that's an argument in favor of having some sort of built-in functionality where you can pass in an environment variable and pare down the tests.

<hr>

Examples of similar approaches you could take:

_Example A_ – specifying test names as an environment variable (the PR also runs test classes/cases that are `focus_`ed, whereas this approach doesn't)

``` objc
// Specify test names as a CSV environment variable

#if INTEGRATION_TESTING
    NSSet *testsToRun;

    NSString *csvTestNames = [[NSProcessInfo processInfo] environment][@"integrationTestNames"];
    if (csvTestNames) {
        NSSet *testNamesToRun = [NSSet setWithArray:[csvTestNames componentsSeparatedByString:@","]];
        testsToRun = [[SLTest allTests] objectsPassingTest:^BOOL(id testKlass, BOOL *stop) {
            return [testNamesToRun containsObject:NSStringFromClass(testKlass)];
        }];
    } else {
        testsToRun = [SLTest allTests];
    }

    [[SLTestController sharedTestController] runTests:testsToRun withCompletionBlock:nil];
#endif
```

<hr>

_Example B_ – grouping tests by a tag

Another approach might be assigning tests a tag to group together test classes, and then filtering by that tag. For example, machine A might run test cases with the tags "login" and "signup", while machine B ran tests tagged "logout" and "send message". This approach wouldn't require updating a config file as often, or with as much data.

``` objc
// Add a category to `SLTest` to get the tag name:

@implementation SLTest (Tags)
+ (NSString *)testTagName
{
    return @"default";
}
@end
```

``` objc
// Subclasses can override the default implementation: 
+ (NSString *)testTagName
{
    return @"DirectMessages";
}
```

``` objc
// Finally, specify test tag names as a CSV environment variable.
#if INTEGRATION_TESTING
    NSSet *testsToRun;

    NSString *csvTestTags = [[NSProcessInfo processInfo] environment][@"integrationTestTags"];
    if (csvTestTags) {
        NSSet *testTagsToRun = [NSSet setWithArray:[csvTestTags componentsSeparatedByString:@","]];
        testsToRun = [[SLTest allTests] objectsPassingTest:^BOOL(id testKlass, BOOL *stop) {
            return [testTagsToRun containsObject:[testKlass testTagName]];
        }];
    } else {
        testsToRun = [SLTest allTests];
    }

    [[SLTestController sharedTestController] runTests:testsToRun withCompletionBlock:nil];
#endif
```

<hr>

_Example C_ – partitioning test cases by number of machines running them

Finally, I can imagine an implementation that passes in _N_, the number of CI machines running the tests and which # the current machine is. After sorting `[SLTest allTests]` by class name, it could partition the tests into _N_ groups, and then run only its portion of the tests.

wearhere:
A couple of devs (Inkling included) have now implemented their own partitioning schemes, so I think there's sufficient demand to build _something_ into Subliminal core. The challenge is identifying a useful lowest-common-denominator. @qiffp's proposal is pretty tempting insofar as test[ case] names are the simplest way to describe the items to run and it's similar to an existing mechanism.

(Though focusing isn't a full partitioning scheme, no? --How do you select the "non-focused" tests to form the second partition, @qiffp? It wouldn't be hard to extend the syntax but doing so would cause CI focusing to diverge from debugging focusing, where you don't care about the second partition.)

I think it's important to clarify whether we'd like partitioning to solve a problem of performance or reporting. My instinct is to say that listing individual test[ case]s, while not really scalable, would be actually be a pretty good way to cut run times, because devs probably have a good sense of the most problematic cases. (Our use case at Inkling, for instance, is to run just one long-running test in its own job, and then all the rest in another.)

@MaxGabriel's tag scheme, by contrast, seems more geared toward grouping tests to produce more meaningful reports. (Though devs could certainly use a tag like "really slow test" :D ) @jzucker2 how are you guys using tags?

**TL;DR** if we want to make runs faster I think we should do name-based whitelisting/blacklisting (i.e. this request, tweaked to form a blacklist too) because it's dead simple and might even produce pretty good performance improvements; and then maybe (in a follow-up request) add partitioning-by-job-number (@MaxGabriel's Example C) 'cause that'll scale better.

qiffp:
@MaxGabriel It's been perfectly scalable for us so far - we haven't actually touched the environment variable setting since the jobs were first set up. This is probably due to the organization of our tests, which are separated such that each test class covers a different section of the app (e.g. one class for settings tests). This way, new tests for new features just get added to these existing test classes.

If we were to separate our test classes by feature, the tag proposal would make perfect sense and the end result would be the same for us (going with the previous example, we'd have multiple test classes, each covering a settings feature and each having a "settings" tag). With our current scheme, it would end up with duplication (the tag would be the same as the test class), but if the majority of Subliminal users are dividing their test classes in smaller sections and frequently creating new test classes, it definitely makes sense to implement this in that way .

I'd actually played around with the third suggestion (give it N machines) prior to writing this, but ended up doing it this way instead because as I imagined it, the N-machines implementation would have led to a lot more complexity and would have had little reporting use. Since the jobs run on different machines and don't communicate, you'd need to supply each job with envvars for N, an identifier for that job to determine which section of the tests it would run (i.e. 1, 2, 3, ..., N) and a randomization seed (so that each job would use the same one, rather than each generating their own), which would then need to be incremented for each run (probably scripted, but another thing for the user to handle).

> (Though focusing isn't a full partitioning scheme, no? --How do you select the "non-focused" tests to form the second partition, @qiffp? It wouldn't be hard to extend the syntax but doing so would cause CI focusing to diverge from debugging focusing, where you don't care about the second partition.)

@wearhere Not sure what you mean by this. We have just divided the tests into the different sections and run a separate job for each of those sections, so each test gets run.

RE: performance vs reporting - In our case, this was written to solve performance/speed issues, but ended up as solving both that and reporting.

wearhere:
@qiffp Let's say that you have tests A, B, C and want to partition that like [A], [B, C]. It sounds like you do that by running

``` sh
# Job 1: Whitelist A
subliminal-test [...] -e FOCUS A
# Job 2: Whitelist B, C
subliminal-test [...] -e FOCUS B,C
```

Is that right?

What I was talking about was a scheme where you could write

``` sh
# Job 1: Whitelist A
subliminal-test [...] -e FOCUS A
# Job 2: Blacklist A (Focus everything _but_ A)
subliminal-test [...] -e FOCUS -A
```

to avoid having to fully enumerate your tests, if you just wanted to pick a couple out to run on their own. The latter scheme is more powerful IMO but confuses the "focus" terminology a bit because currently (when prefixing test[ case] names) it's only used to form a whitelist--so, if we wanted to enable such a scheme, maybe we'd want to use a different env variable than "FOCUS".

wearhere:
@qiffp, after talking with some other users and revisiting earlier attempts along these lines, I've implemented a more general purpose, tag-based solution in the form of #233. It can be used for this purpose too, though, because tests and test cases are automatically tagged with their names: `subliminal-test [...] -e SL_TAGS testFoo,testBar` should be equivalent to your `subliminal-test [...] -e FOCUS testFoo,testBar`. I'd love to know what you think.

qiffp:
@wearhere, closing this now since you've covered it in #233

wearhere:
Thanks again for the proposal @qiffp. The discussion here (h/t @MaxGabriel too) really helped me clarify how such a feature might work.

Issue #201, opened 2014-06-03
SLTextField isn't tappable


justinseanmartin:
I was debugging an issue yesterday and I believe it is the same as described here:

http://stackoverflow.com/questions/23838808/how-can-i-make-an-sltextfield-tappable

I debugged it for a while, did some playing around with the UIAutomation JavaScript being generated, and found a patch that fixed the issue.  The accessibility hierarchy being generated was:

UIWindow -> STTableView (our subclass) -> UITableViewCellScrollView -> SQTextField (our subclass)

However, it seems like the accessibility hierarchy that worked from instruments was:

UIWindow -> STTableView -> SQTextField

I have a patch that fixes the behavior I was seeing, but I'm not sure what other side effects it might have (otherwise I'd just open a PR). Also, I can't tap on the text controls in iOS6 (with or without my patch), so I excluded modifying the behavior on that platform.

Patch: https://github.com/justinseanmartin/Subliminal/commit/e8cc1e5e2d0d92d38f7f232120eafa3a08d11091

justinseanmartin:
I've opened a PR with a corresponding integration test case: https://github.com/inkling/Subliminal/pull/202

wearhere:
Did #202 resolve this @justinseanmartin?

justinseanmartin:
Yes, this issue is now fixed in master.

wearhere:
:thumbsup:

PR #202, opened 2014-06-04
Fix UITextView accessibility path when contained within a UITableViewCell

justinseanmartin:
In iOS7, having a UITextView within a UITableViewCell doesn't appear to work as expected. The fix appears to function as expected. 

I haven't fixed the second test case that I added, and clear current text doesn't appear to work on either iOS 6 or iOS 7. I'll be debugging this issue now.

### DIFF
--- a/Integration Tests/Tests/SLTextFieldTest.m
+++ b/Integration Tests/Tests/SLTextFieldTest.m
@@ -40,8 +40,10 @@
     [super setUpTestCaseWithSelector:testSelector];
 
     if (testSelector == @selector(testSetText) ||
+        testSelector == @selector(testSetTextWithinTableViewCell) ||
         testSelector == @selector(testSetTextCanHandleTapHoldCharacters) ||
         testSelector == @selector(testSetTextClearsCurrentText) ||
+        testSelector == @selector(testSetTextClearsCurrentTextWithinTableViewCell) ||
         testSelector == @selector(testSetTextWhenFieldClearsOnBeginEditing) ||
         testSelector == @selector(testGetText) ||
         testSelector == @selector(testDoNotMatchEditorAccessibilityObjects) ||
@@ -68,6 +70,12 @@
     SLAssertTrue([SLAskApp(text) isEqualToString:expectedText], @"Text was not set to expected value.");
 }
 
+- (void)testSetTextWithinTableViewCell {
+    NSString *const expectedText = @"foo";
+    SLAssertNoThrow([UIAElement(_textField) setText:expectedText], @"Should not have thrown.");
+    SLAssertTrue([SLAskApp(text) isEqualToString:expectedText], @"Text was not set to expected value.");
+}
+
 - (void)testSetTextCanHandleTapHoldCharacters {
     NSString *const expectedText = @"foo’s a difficult string to type!";
     SLAssertNoThrow([UIAElement(_textField) setText:expectedText], @"Should not have thrown.");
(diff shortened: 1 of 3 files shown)


justinseanmartin:
Ok, so my fix to the issue described in https://github.com/inkling/Subliminal/issues/201 doesn't work quite right.  The UITextView is showing up as the UITableViewCell in the hierarchy, and the actual UITextView element doesn't appear in the accessibility path. 

This is why it seemed like it was working. Tapping on the UITableViewCell behaves like tapping on the text view. However, you can't send setValue('') on the cell and have it expect to work properly.

I'll need to look at my fix some more, or any pointers would be great.

justinseanmartin:
The latest change gets it working on iOS 7, but iOS 6 still doesn't work.

wearhere:
Haha I was just about to post a response! Great debugging--the problem is indeed that there's a missing element in the path. That missing element looks like it's the text field, but it's actually the table view cell: `UITableViewCell` syncs its accessibility properties with those of its children. I found this out by looking at the path that UIAutomation generated when recording a tap on the text field, vs. the filtered accessibility element path generated for the text field (e.g. breaking at `SLAccessibilityPath.m:403`, then printing _filteredAccessibilityElementPath_).

Here's the problem/fix:

``` diff
diff --git a/Sources/Classes/UIAutomation/User Interface Elements/NSObject+SLAccessibilityHierarchy.m b/Sources/Classes/UIAutomation/User Interface Elements/NSObject+SLAccessibilityHierarchy.m
index 79310e9..c25799a 100644
--- a/Sources/Classes/UIAutomation/User Interface Elements/NSObject+SLAccessibilityHierarchy.m  
+++ b/Sources/Classes/UIAutomation/User Interface Elements/NSObject+SLAccessibilityHierarchy.m  
@@ -373,6 +373,9 @@


 @implementation UITableViewCell (SLAccessibilityHierarchy)
+- (BOOL)classForcesPresenceInAccessibilityHierarchy {
+    return YES;
+}
 - (BOOL)classForcesPresenceOfMockingViewsInAccessibilityHierarchy {
     return YES;
 }
```

Before, we assumed that `UITableViewCells` themselves wouldn't appear in the accessibility hierarchy, only the artificial elements that UIAccessibility generates to mock the cells' properties. I guess that's not always true! Hat-tip to @jzucker2 for this solution, it's drawn from c129333b43eef18cd7639fd4e185c9a26ff03acc.

wearhere:
The "field elements" are a red herring, btw. UIAutomation does include them in its recorded path:

``` js
// all with the label "test element": cell, text field, field element
target.frontMostApp().mainWindow().tableViews()["Empty list"].cells()["test element"].textFields()["test element"].textFields()["test element"].setValue('');
```

But they're not actually necessary--you can remove the last `.textFields()["test element"]` and that line will evaluate fine--besides which they don't exist on iOS 6 as you saw. Sometimes, they're not recognized by UIAutomation at all: `-[SLTextFieldTest testDoNotMatchEditorAccessibilityObjects]` verifies that we _don't_ try to match them.

wearhere:
If you remove 6e8796b and add the patch above to this PR, I'll merge--it looks great otherwise.

justinseanmartin:
Keep the part from e8cc1e5?  It doesn't seem to be necessary to get iOS7 working with 6e8796b, so I removed it.

justinseanmartin:
Nevermind, iOS7 doesn't work without e8cc1e5. Putting that back in and we should have a clean run.

justinseanmartin:
Looks like one Travis timing issue failure on iOS 6 iPhone:

ElementMatchingTest.m:124: "actualWaitTimeInterval - expectedWaitTimeInterval < [self waitDelayVariability]" should be true. Test waited for 2.70998 but should not have waited appreciably longer than 2.

```
    ✗ "testElementsWaitToMatchValidObjects" failed.
```

I would presume this isn't caused by my changes however.

wearhere:
I'll merge this once #205 goes in (see my comment on #204).

wearhere:
Thanks @justinseanmartin!

PR #203, opened 2014-06-05
Add support for checking existence of labels

justinseanmartin:
I'd like to check that a progress spinner is appearing and subsequently disappears as I'm transitioning through screen states. At first, I didn't realize that I could just use SLElement to match any element type, as it doesn't match the way I would use UIAutomation directly. Even still, I think it makes sense to have the type checking in 'matchesObject'. Thoughts?

Marking this as WIP, as I really should add an integration test before merging.


Sources/Classes/UIAutomation/User Interface Elements/SLLabel.h
```diff
@@ -0,0 +1,32 @@
...
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+#import "SLElement.h"
+
+/**
+ `SLWindow` matches instances of `UILabel`.
```
MaxGabriel: `SLLabel`

Sources/Classes/UIAutomation/User Interface Elements/SLLabel.h
```diff
@@ -0,0 +1,32 @@
...
+ 
+ They generally can't be interacted with, but you can ensure that they appear on the screen at least.
```
MaxGabriel: IMO this line isn't likely to add information that the reader doesn't already know

Sources/Classes/UIAutomation/User Interface Elements/SLLabel.h
```diff
@@ -0,0 +1,32 @@
...
+ */
```
MaxGabriel: I think its worth mentioning that `UILabel`s automatically inherit their accessibility label from their `text` value.

As an implication of this, maybe mention that you can used the `accessibilityIdentifier` property if you want to match a label, but don't know what its text will be at runtime (This is more general advice so not sure if it fits here).

Sources/Classes/UIAutomation/User Interface Elements/SLLabel.h
wearhere: Agreed.

Sources/Classes/UIAutomation/User Interface Elements/SLLabel.h
wearhere: Also, what do you think about matching objects with the `UIAccessibilityTraitStaticText` trait instead of specifically `UILabel`, @justinseanmartin? That's matching more on behavior rather than implementation and would allow matching e.g. "labels" in web views too.

wearhere: With something like the comment: "Such as, but not limited to, instances of `UILabel`." (See `SLButton`'s overview.)

Sources/Classes/UIAutomation/User Interface Elements/SLLabel.h
wearhere: I think that both points are actually general advice regarding `SLElement`. The documentation of `SLElement`'s constructors link to the [UIAccessibility Protocol Reference](https://developer.apple.com/library/ios/documentation/uikit/reference/UIAccessibility_Protocol/Introduction/Introduction.html) to provide general guidance around how labels might be derived or set.

Besides those notes, it might be worth adding a general "how to match elements" guide to the wiki somewhere.

The `accessibilityIdentifier` suggestion is definitely worth adding to `+[SLElement elementWithAccessibilityIdentifier:]` and/or the wiki.

Filed #206 to cover these topics.

wearhere:
In summary of the discussion here @justinseanmartin, I like the idea of `SLLabel` but think it probably should match elements with the `UIAccessibilityTraitStaticText` trait instead of just `UILabel`. Also please do add an integration test (`SLButtonTest` is a similar example).

Sources/Classes/UIAutomation/User Interface Elements/SLLabel.h
justinseanmartin: Removed.

Sources/Classes/UIAutomation/User Interface Elements/SLLabel.h
justinseanmartin: What about OR'ing the two conditions? That way if someone happened to subclass UILabel and made it non-static, that would still work as well. Granted, it would be a bit weird for someone to do...

justinseanmartin:
Thanks @MaxGabriel & @wearhere - All feedback should be incorporated now, including 2 new integration tests.

wearhere:
> What about OR'ing the two conditions? That way if someone happened to subclass UILabel and made it non-static, that would still work as well. Granted, it would be a bit weird for someone to do...

You know, I think we should avoid the question of whether labels should be static by just making this class called `SLStaticText`, matching elements with the `UIAccessibilityTraitStaticText` trait. This keeps the framework's focus on what the element does (for the user), rather than on how it's implemented underneath; and will make it more sensical for this class to match both labels and web text--where the latter isn't necessarily a "label" in the context of the web page. UIAutomation takes the same tack by having `UIAStaticText`, but not `UIALabel`.

(Granted, "static text" doesn't immediately suggest "label"--I think the class description should say something like "`SLStaticText` matches objects that have the `UIAccessibilityTraitStaticText` accessibility trait, like labels and text on web pages.")

If you did really want to match a label--though the tests shouldn't care about this IMO--you could do:

``` objc
#import <Subliminal/SLUIAElement+Subclassing.h>

// in a test
SLElement *label = [SLElement elementMatching:^(NSObject *object) {
    return [object isKindOfClass:[UILabel class]] && /* some other criteria */;
}];
```

Btw, you have to remove the focus on the tests before committing. Thanks!

justinseanmartin:
Ok, sounds good on SLStaticText, will make that change now. 

Also, _headslap_ on the focus_...

wearhere:
:+1:

PR #204, opened 2014-06-05
Allow Class objects to be passed to selectors

justinseanmartin:
The Class object conforms to NSCopying, so there is no reason they shouldn't be able to be passed as a parameter to a selector. I was thinking about using this to see if the current main windows rootViewController matched the one I was expecting. I'm rethinking the approach, but I don't see why it shouldn't at least be possible.


justinseanmartin:
This Travis failure is a different test case from the one seen in https://github.com/inkling/Subliminal/pull/202, but it looks like the same premise of being a timing issue. in 'testWaitUntilTappableYESThenPerformActionWithUIARepresentationWaitsUntilTappable'. Should these tests be removed or altered somehow?

wearhere:
Yeah, we needed to allow extra time for Travis, basically: #205. That should fix both the failure seen here and that which was seen in #202. I'm fine merging this given that the unit tests passed, but I'm going to wait to do so until #205 is merged so that the tests don't fail again.

PR #209, opened 2014-06-11
Disable integration test transition animations when running in CI.

wearhere:
There's no one watching in that case, and disabling animations speeds up the tests
by about 40%.


wearhere:
Note that this request [already passed Travis](https://travis-ci.org/inkling/Subliminal/builds/28358749)--I just rebased to fix up a commit message.

PR #211, opened 2014-06-14
Add support for UIPickerViews in Subliminal

justinseanmartin:
UIPickerViews are under a different Window other than keyWindow, so I had to find the other one.  Otherwise it is pretty straightforward.  I'm leaning on the UIA methods a bit more heavily for interacting with the element, but open to suggestions.

Added one fairly comprehensive test case.


Sources/Classes/UIAutomation/User Interface Elements/SLPickerView.m
```diff
@@ -0,0 +1,70 @@
...
+    
+    return nil;
+}
+
+- (int)numberOfComponentsInPickerView {
+    return [(NSString *)[self waitUntilTappable:YES thenSendMessage:@"wheels().length"] intValue];
+}
+
+- (NSArray *)valueOfPickerComponents {
+    __block NSArray *pickerComponenValues;
```
MaxGabriel: Typo here, missing the t on component

Integration Tests/Tests/SLPickerViewTestViewController.m
```diff
@@ -0,0 +1,75 @@
...
+
+- (instancetype)initWithCoder:(NSCoder *)coder
+{
+    self = [super initWithCoder:coder];
+
+    if (self) {
+        _pickerView = [[UIPickerView alloc] init];
+        _pickerView.delegate = self;
+        _pickerView.accessibilityIdentifier = @"Picker View";
+        _values = [[NSMutableArray alloc] initWithObjects:[[NSNumber alloc] initWithInt:1], [[NSNumber alloc] initWithInt:1], nil];
```
MaxGabriel: It'd be nicer to use literals for the `NSNumber`s here (`@1`)

Integration Tests/Tests/SLPickerViewTestViewController.m
```diff
@@ -0,0 +1,75 @@
...
+- (NSInteger)numberOfComponentsInPickerView:(UIPickerView *)pickerView {
+    return 2;
+}
+
+- (NSInteger)pickerView:(UIPickerView *)pickerView numberOfRowsInComponent:(NSInteger)component {
+    return 3;
+}
+
+- (void)pickerView:(UIPickerView *)pickerView didSelectRow:(NSInteger)row inComponent:(NSInteger)component {
+    [_values setObject:[[NSNumber alloc] initWithInt:(row + 1)] atIndexedSubscript:component];
```
MaxGabriel: You could do `@(row + 1)` instead

Sources/Classes/UIAutomation/User Interface Elements/SLPickerView.h
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ `SLPickerView` needs to find another window other than the mainWindow 
+ to search.
+ */
+@interface SLPickerView : SLElement
+
+- (int)numberOfComponentsInPickerView;
+- (NSArray *)valueOfPickerComponents;
+- (BOOL)selectValue:(NSString *)value forComponent:(int)wheelNumber;
```
MaxGabriel: Can you document these? Its super nice for library users since the docs show up in autocomplete info and the sidebar + Travis fails without it https://travis-ci.org/inkling/Subliminal/jobs/27543975

Integration Tests/Tests/SLPickerViewTestViewController.m
MaxGabriel: ooh, actually you could do `_values[component] = @(row + 1)`

Integration Tests/Tests/SLPickerViewTestViewController.m
```diff
@@ -0,0 +1,75 @@
...
+//  Copyright (c) 2014 Inkling. All rights reserved.
+//
+
+#import "SLTestCaseViewController.h"
+
+#import <Subliminal/SLTestController+AppHooks.h>
+
+@interface SLTestPickerTextField : UITextField <UIPickerViewDelegate, UIPickerViewDataSource>
+@property UIPickerView *pickerView;
+@end
```
MaxGabriel, line 14: Any reason to subclass UITextField and have it be a delegate for the picker? I guess I could envision a specialized `DateTextField` or something that managed its own picker, but in this case it just seemed unusual when compared to using the controller.

Integration Tests/Tests/SLPickerViewTest.m
```diff
@@ -0,0 +1,72 @@
...
+    SLAssertTrue([[UIAElement(textField) value] isEqualToString:@"2 - 1"], @"SLTextField should have updated to match the UITextField.");
+
+    // Select value 1 on the first picker component
+    SLAssertTrue([UIAElement(pickerView) selectValue:@"3" forComponent:1], @"SLPickerView should be able to set the first value.");
+    values = [UIAElement(pickerView) valueOfPickerComponents];
+    SLAssertTrue([values count] == 2, @"SLPickerView should have 2 spinner components. Actual: %d", [values count]);
+    SLAssertTrue([[values objectAtIndex:0] isEqualToString:@"2. 2 of 3"], @"SLPickerView first spinner should have value '2. 1 of 3'. Actual: %@", [values objectAtIndex:0]);
+    SLAssertTrue([[values objectAtIndex:1] isEqualToString:@"3. 3 of 3"], @"SLPickerView second spinner should have value '3. 3 of 3'. Actual: %@", [values objectAtIndex:1]);
+    SLAssertTrue([[UIAElement(textField) value] isEqualToString:@"2 - 3"], @"SLTextField should have updated to match the UITextField.");
+}
```
MaxGabriel, line 79: This should probably be split out into smaller test cases. When all the assertions together are in one method, its easier for the individual assertions to interfere with one another (e.g. an earlier test case might do something innocuous that breaks a later test case).

Additionally, because the test case names are used to report the test results, you end up with better, more accurate output as to what tests failed and passed. Finally, splitting up into separate methods lets you run an individual one using `focus_`, which would be nice if you were debugging just that one.

Integration Tests/Tests/SLPickerViewTestViewController.m
```diff
@@ -0,0 +1,75 @@
...
+    [self updateText];
+}
+
+- (void)updateText {
+    self.text = [_values componentsJoinedByString:@" - "];
+}
+
+- (CGRect)caretRectForPosition:(UITextPosition *)position {
+    return CGRectZero;
+}
```
MaxGabriel: Is this doing anything?

Sources/Classes/UIAutomation/User Interface Elements/SLPickerView.m
justinseanmartin: Fixed in next drop.

Integration Tests/Tests/SLPickerViewTestViewController.m
justinseanmartin: Fixed in next drop.

Integration Tests/Tests/SLPickerViewTestViewController.m
justinseanmartin: Fixed in next drop.

Sources/Classes/UIAutomation/User Interface Elements/SLPickerView.h
justinseanmartin: Fixed in next drop.

Integration Tests/Tests/SLPickerViewTest.m
justinseanmartin, line 79: Agreed, will get on this.

Integration Tests/Tests/SLPickerViewTestViewController.m
justinseanmartin: The only thing it does is make the cursor not blink in the text box. Again, it was more just matching our use case, but it probably doesn't make any difference to the actual test itself. Happy to pull it out or just comment it as desired.

Integration Tests/Tests/SLPickerViewTestViewController.m
justinseanmartin, line 14: I only did it this way, because it is the use case where we're using it in our app, so I figured I'd validate our scenario was working. Presumably it could be hooked up to any control.

I actually just spent some time playing with the control some more. It can either be hosted in the view itself like any other view element, but it is different when it is the inputView for a textField control. Specifically, it isn't in a different window, which makes sense. This means that the SLPickerView needs to handle both cases however.

I should probably validate that the UIDatePicker variant is works as well.

Integration Tests/Tests/SLPickerViewTestViewController.m
justinseanmartin: Added a comment for now, but can remove if desired.

Integration Tests/Tests/SLPickerViewTest.m
```diff
@@ -0,0 +1,72 @@
...
+    SLPickerView *pickerView = [SLPickerView elementWithAccessibilityIdentifier:@"Picker View"];
+
+    SLAssertTrue([[UIAElement(textField) value] isEqualToString:@"1 - 1"], @"SLTextField should have matched the UITextField.");
+    SLAssertFalse([pickerView isValidAndVisible], @"The Picker View shouldn't be visible initially");
+
+    // Tapping the text field should make the picker view appear
+    SLAssertNoThrow([UIAElement(textField) tap], @"SLTextField should be tappable.");
+    SLAssertTrue([UIAElement(pickerView) isValid], @"SLPickerView should be valid.");
+    // NOTE: The element is not considered visible, because it is fully covered by its spinners
+    SLAssertFalse([UIAElement(pickerView) isVisible], @"SLPickerView itslef isn't visible.");
```
wearhere: This was a reasonable assumption but this is actually a bug in `-[SLElement isVisible]` :) Now that we match elements in non-key windows, [this line](https://github.com/inkling/Subliminal/blob/master/Sources/Classes/Internal/NSObject+SLVisibility.m#L279) just needs to be changed to read `self.window` and then the picker will be reported as visible.

Would make a good test case to add to `SLElementVisibilityTest`, too.

wearhere: Actually a proper fix is slightly more complicated than that: https://github.com/inkling/Subliminal/issues/212. Can you leave the assertion as is, but add a "to fix" comment with a link to #212?

justinseanmartin: I've temporarily worked around the issue by writing a custom isVisible method that relies on the UIA implementation for this element.  It might not be perfect, but it is good enough to get me unblocked for now.

Integration Tests/Tests/SLPickerViewTest.m
justinseanmartin, line 79: Done in the latest commit.

Integration Tests/Tests/SLPickerViewTestViewController.m
justinseanmartin, line 14: Added support for UIDatePicker and verified the functionality works both for elements hosted within a inputView as well as in the main window.

justinseanmartin:
Latest commit should have handled all PR feedback and is pretty sizable.  I don't love a few things about the implementation, but it gets the job done. Suggestions welcome.

The things I felt aren't great solutions are:
- the way SLDatePicker subclasses SLPickerView (overrides on classToMatchOn & wheelsObjectPathInUIA in SLPickerView+Internal.h)
- the override of isTappable in SLPickerView

Sources/Classes/UIAutomation/User Interface Elements/SLDatePicker.h
```diff
@@ -0,0 +1,43 @@
...
+//  limitations under the License.
+//
+
+#import "SLPickerView.h"
+
+/**
+ SLDatePicker elements represent instances of UIDatePicker, which aren't subclassed
+ from UIPickerView. This object gives access to viewing and manipulating the individual
+ components (or as UIA calls them 'wheels') of the Picker. UIA represents the controls
+ in almost the exact same way, so we treat it as a subclass of SLPickerView.
```
wearhere: What if we made `SLDatePicker` re-implement `SLPickerView`'s interface without subclassing? Since it matches a different type of object. It could forward the interface to an internal `SLPickerView`. Here's a patch showing how that could work: http://cl.ly/code/1F2V0i3J1L1n.

Sources/Classes/UIAutomation/User Interface Elements/SLDatePicker.h
```diff
@@ -0,0 +1,43 @@
...
+
+ The only difference is that the UIDatePicker element shows up in the accessibility
+ hierarchy as a UIAElement rather than a UIAPicker. The UIAPicker is actually a child
+ element, which is backed by a UIView object with no accessibility information on it.
+ Ideally, we'd match on the UIView within the UIDatePicker, but that seems harder to 
+ make work given the existing infrastructure.
```
wearhere: I'm not sure the user needs to know this stuff--it seems kinda like implementation details.

justinseanmartin: Yea, I was going on the verbose side, but I didn't really split what should be in the appledoc versus what should be in the implementation comments. I'll scrub this.

Sources/Classes/UIAutomation/User Interface Elements/SLElement.m
```diff
@@ -178,13 +178,19 @@ - (void)setShouldDoubleCheckValidity:(BOOL)shouldDoubleCheckValidity {
...
+}
+
 - (SLAccessibilityPath *)accessibilityPathWithTimeout:(NSTimeInterval)timeout {
     __block SLAccessibilityPath *accessibilityPath = nil;
     NSDate *startDate = [NSDate date];
     // a timeout of 0 means check once--but then return immediately, no waiting
     do {
         dispatch_sync(dispatch_get_main_queue(), ^{
-            accessibilityPath = [[[UIApplication sharedApplication] keyWindow] slAccessibilityPathToElement:self];
+            accessibilityPath = [[self accessibilityPathSearchRootElement] slAccessibilityPathToElement:self];
```
wearhere: This is a clever way to avoid searching multiple windows but I think it might be alright to just do that. We probably don't have to search through all the windows, just the key window and those above it:

``` diff
-            accessibilityPath = [[self accessibilityPathSearchRootElement] slAccessibilityPathToElement:self];
+            NSArray *windows = [[UIApplication sharedApplication] windows];
+            UIWindow *keyWindow = [[UIApplication sharedApplication] keyWindow];
+            NSUInteger keyWindowIndex = [windows indexOfObject:keyWindow];
+            NSAssert(keyWindowIndex != NSNotFound,
+                     @"`[[UIApplication sharedApplication] windows]` should contain the key window.");
+            
+            for (NSUInteger windowIndex = keyWindowIndex; windowIndex < [windows count]; windowIndex++) {
+                UIWindow *window = windows[windowIndex];
+                accessibilityPath = [window slAccessibilityPathToElement:self];
+                if (accessibilityPath) break;
+            }
```

And then elements wouldn't require any special configuration depending on where they displayed.

Sources/Classes/UIAutomation/User Interface Elements/SLDatePicker.h
wearhere: Oh yeah I like to err on the side of commenting in the implementation, given that the accessibility hierarchy is rarely intuitive ;) , just not in the header.

Sources/Classes/UIAutomation/User Interface Elements/SLPickerView.h
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ SLPickerView elements represent instances of UIPickerView.
+
+ If it is the inputView for a text field, set isPickerForTextInputView to YES and it will
+ be contained looked for in the 'UITextEffectsWindow' window rather than the keyWindow.
+
+ This object gives access to viewing and manipulating the individual components (or as UIA
+ calls them 'wheels') of the Picker.
+ */
```
wearhere, line 32: Let's mention that `SLDatePicker` should be used to manipulate date pickers, 'cause there's no way to specify the accessibility information on a `UIDatePicker`'s picker view.

Sources/Classes/UIAutomation/User Interface Elements/SLPickerView.h
```diff
@@ -0,0 +1,76 @@
...
+@interface SLPickerView : SLElement
+
+/**
+ Calls the equivalent of 'wheels().length' on the picker in UIA. This will return
```
wearhere: It's nice to retain UIAutomation concepts where possible, i.e. your reference to "wheels", but let's avoid going so far as to provide equivalent UIAutomation code--that strikes me as too much information.

Sources/Classes/UIAutomation/User Interface Elements/SLPickerView.m
```diff
@@ -0,0 +1,108 @@
...
+        NSString *responseString = [[SLTerminal sharedTerminal] evalWithFormat:
+                                       @"var values = [];\n"
+                                        "var wheels = %@.%@;\n"
+                                        "for (var i = 0; i < wheels.length; i++) {\n"
+                                        "    values.push(wheels[i].value());\n"
+                                        "}\n"
+                                        "JSON.stringify(values);",
+                                        uiaRepresentation, [self wheelsObjectPathInUIA]];
+        NSData *jsonDataFromString = [responseString dataUsingEncoding:NSUTF8StringEncoding];
+        pickerComponentValues =  (NSArray *)[NSJSONSerialization JSONObjectWithData:jsonDataFromString
```
wearhere, line 59: `NSJSONSerialization` will throw an exception if invoked with `nil` data. Let's make our potential failure ;D a little more explicit:

``` objc
NSData *jsonDataFromString = [responseString dataUsingEncoding:NSUTF8StringEncoding];
NSAssert(jsonDataFromString, @"%s JavaScript is malformed.", __PRETTY_FUNCTION__);
```

Integration Tests/Tests/SLDatePickerTest.m
```diff
@@ -0,0 +1,70 @@
...
+- (void)setUpTestCaseWithSelector:(SEL)testSelector {
+    [super setUpTestCaseWithSelector:testSelector];
+    
+    _pickerView = [SLDatePicker elementWithAccessibilityIdentifier:@"Picker View"];
+    
+}
+
+#pragma mark - Test Cases
+
+- (void)testDatePickerCanBeFoundAfterTappingText {
```
wearhere: These first two test cases seem to be copypasta from the regular picker test :) , there's no text field in this test.

justinseanmartin: Yep, you got me. Will fix naming. I still think having the test cases for finding both inside and outside of the textField's inputView is useful.

wearhere: Definitely. Though you just reminded me, I think it would be nicest to have tests of `SLPickerView` and `SLDatePicker` where we just test them as regular views.

The case where the picker is an input view exercises the picker view itself less than Subliminal's ability to match elements in other windows (especially if we make [this change](https://github.com/inkling/Subliminal/pull/211/files#r13839036))--I think the input view scenario could be an `SLElementMatchingTest` case, actually.

Integration Tests/Tests/SLDatePickerTest.m
```diff
@@ -0,0 +1,70 @@
...
+}
+
+- (void)testDatePickerReportsCorrectNumberOfComponents {
+    SLAssertTrue([UIAElement(_pickerView) numberOfComponentsInPickerView] == 3, @"SLDatePicker should have 3 components by default.");
+}
+
+// Default value is hardcoded to 12/12/2012
+- (void)testDatePickerInitialComponentValues {
+    NSArray *values = [UIAElement(_pickerView) valueOfPickerComponents];
+    SLAssertTrue([values count] == 3, @"SLDatePicker should have 3 spinner components. Actual: %d", [values count]);
```
wearhere: For Subliminal to build for 64-bit targets, we can't use "%d" or "%u"--instead, we should use:
- `"%lu", (unsigned long)foo` when `foo` is an `NSUInteger`, and
- `"%li", (long)foo` when `foo` is an `NSInteger`

Sources/Classes/UIAutomation/User Interface Elements/SLElement.m
justinseanmartin: Taken verbatim.

Sources/Classes/UIAutomation/User Interface Elements/SLDatePicker.h
justinseanmartin: I like it, also picked this up verbatim.

Sources/Classes/UIAutomation/User Interface Elements/SLPickerView.m
justinseanmartin, line 59: Done.

Sources/Classes/UIAutomation/User Interface Elements/SLPickerView.m
```diff
@@ -0,0 +1,108 @@
...
+// We should be waiting until the element is tappable before setting the value, but it doesn't
+// fit with the existing infrastructure to get the waitUntilTappable:thenSendMessage: method
+// to use an overloaded isTappable method.
+- (void)selectValue:(NSString *)title forComponent:(int)componentIndex {
+    [self waitUntilTappable:NO thenSendMessage:@"%@[%d].selectValue(\"%@\")",
+                                    [self wheelsObjectPathInUIA], componentIndex, title];
+}
+
+// Overload behavior of SLUIAElement, because testTextInputPickerViewCanBeFoundAfterTappingText
+// fails saying that the element isn't tappable on iOS 6.
```
wearhere, line 76: Oh, UIAutomation. Filed a follow-up ticket (#214) to improve this somewhat.

wearhere:
Ok @justinseanmartin, back to you. Thanks for working through this, this has turned out to be a surprisingly intricate addition!

wearhere:
And btw @MaxGabriel, your comments on the first rev of the PR were great. You are more than welcome to continue to chime in if you see anything that might be improved.

Integration Tests/Tests/SLPickerViewTest.m
wearhere: Oh yeah, falling back to `UIAElement.isVisible` is a great temporary fix.

Integration Tests/Tests/SLDatePickerTest.m
justinseanmartin: I suppose I should stop directly using int as well and always refer to them with NSInteger and NSUInteger?

Sources/Classes/UIAutomation/User Interface Elements/SLPickerView.h
justinseanmartin, line 32: Not sure I follow the "no way to specific accessibility information", but I have added "To manipulate a date picker control, use SLDatePicker as it isn't a direct subclass
 of UIPickerView."  Is that sufficient?

Sources/Classes/UIAutomation/User Interface Elements/SLPickerView.h
justinseanmartin: Ok, pulled those out for all the SLPickerView comments. Thought it would be nice to understand for someone who knew it from a UIA perspective, but I get the concern. It should be irrelevant (and pretty obvious) as to how it is implemented.

Sources/Classes/UIAutomation/User Interface Elements/SLPickerView.h
wearhere, line 32: I'd say "as `UIDatePicker` isn't a direct subclass of `UIPickerView`" but yes that's a simpler way of putting what I was trying to say.

Integration Tests/Tests/SLDatePickerTest.m
wearhere: Ah, yes please.

Integration Tests/Tests/SLDatePickerTest.m
justinseanmartin: Done. Added one SLElementMatchingTest for UITextField InputViews and the rest are all testing in a normal view element.

justinseanmartin:
All feedback should now be incorporated.

Integration Tests/Tests/SLTextFieldPickerTest.m
```diff
@@ -0,0 +1,81 @@
...
+//  Copyright (c) 2014 Inkling. All rights reserved.
+//
+
+#import "SLIntegrationTest.h"
+
+@interface SLTextFieldPickerViewTest : SLIntegrationTest
+@end
+
+// All of these test cases focus on containing a UIPickerView in a TextField's inputView
+@implementation SLTextFieldPickerViewTest {
```
wearhere: I think the comment/test name (and the test case names below) are out of date now that the input view scenario has moved to `SLElementMatchingTest`.

Integration Tests/Tests/SLTextFieldPickerTest.m
```diff
@@ -0,0 +1,81 @@
...
+    SLAssertTrue([[values objectAtIndex:0] isEqualToString:@"2. 2 of 3"], @"SLPickerView first spinner should have value '2. 1 of 3'. Actual: %@", [values objectAtIndex:0]);
+    SLAssertTrue([[values objectAtIndex:1] isEqualToString:@"3. 3 of 3"], @"SLPickerView second spinner should have value '3. 3 of 3'. Actual: %@", [values objectAtIndex:1]);
+}
+
+- (void)testTextInputPickerViewValuesArentModifiedOnInvalidSelections {
+    // Get us in a good state showing "3 - 1"
+    [UIAElement(_pickerView) selectValue:@"3" forComponent:0];
+
+    // Select an invalid value on the second picker component
+    SLAssertThrowsNamed([UIAElement(_pickerView) selectValue:@"0" forComponent:1], @"SLUIAElementAutomationException", @"SLPickerView should return false on trying to select an invalid value.");
```
wearhere: Please use the actual `SLUIAElementAutomationException` constant for safety.

wearhere: Also it seems like the comment should say "should throw..." vs. "should return false".

Sources/Classes/Internal/SLPickerViewOverrides.h
```diff
@@ -0,0 +1,52 @@
+//
```
wearhere: I don't think this file is necessary anymore.

Sources/Classes/UIAutomation/User Interface Elements/SLDatePicker.h
```diff
@@ -0,0 +1,59 @@
...
+ to viewing and manipulating the individual components (or as UIA calls them 'wheels')
+ of the Picker.
+ */
+@interface SLDatePicker : SLElement
+
+/**
+ Fetches the number of components from UIA.
+ 
+ @return The actual number of components visible on the screen
+ */
```
wearhere, line 39: Methods that talk to UIAutomation (using the `-waitUntilTappable:...` methods) need to document that some standard exceptions may be thrown in the process. They always need to document that `SLUIAElementInvalidException` may be thrown. If they wait until tappable, they need to document that `SLUIAElementNotTappableException` may be thrown too. (This would apply to `-selectValue:forComponent:` except for #214.) You can copy the boilerplate for this from `-[SLTextField setText:withKeyboard:]`.

Sources/Classes/UIAutomation/User Interface Elements/SLDatePicker.h
```diff
@@ -0,0 +1,59 @@
...
+ @return An array of strings containing the value for each wheel component. The string
+ values are of the format "%@ (%d of %d)", where the first portion is current selected
+ wheel's value, and the two numbers represent the current selected element row and the
+ total number of rows.
+ */
+- (NSArray *)valueOfPickerComponents;
+
+/**
+ Changes the selected value of a given component to a specified value on the wheel. If
+ invalid values are passed, an exception will be thrown.
```
wearhere: Rather than documenting the exception in the method discussion like this, please use the "@exception" syntax:

```
@exception SLUIAElementAutomationException Raised if _title_ is not a valid value for the specified component.
```

Sources/Classes/UIAutomation/User Interface Elements/SLElement.m
```diff
@@ -184,7 +184,17 @@ - (SLAccessibilityPath *)accessibilityPathWithTimeout:(NSTimeInterval)timeout {
     // a timeout of 0 means check once--but then return immediately, no waiting
     do {
         dispatch_sync(dispatch_get_main_queue(), ^{
-            accessibilityPath = [[[UIApplication sharedApplication] keyWindow] slAccessibilityPathToElement:self];
+            NSArray *windows = [[UIApplication sharedApplication] windows];
```
wearhere, line 191: Can you put a comment describing what we're doing and why?

wearhere, line 191: (searching for the element in the key window and the windows above it i.e. the text effects window; as a heuristic, to ignore hidden windows)

Sources/Classes/UIAutomation/User Interface Elements/SLPickerView.h
```diff
@@ -0,0 +1,62 @@
...
+//
+
+#import "SLElement.h"
+
+/**
+ SLPickerView elements represent instances of UIPickerView. This object gives access
+ to viewing and manipulating the individual components (or as UIA calls them 'wheels') 
+ of the Picker.
+ 
+ To manipulate a date picker control, use UIDatePicker as it isn't a direct subclass
```
wearhere: "use `SLDatePicker`, as `UIDatePicker` isn't a subclass of `UIPickerView`." I think I was unclear in chat :)

Sources/Classes/UIAutomation/User Interface Elements/SLPickerView.h
```diff
@@ -0,0 +1,62 @@
...
+ of UIPickerView.
+ */
+@interface SLPickerView : SLElement
+
+/**
+ Fetches the number of components from UIA.
+
+ @return The actual number of components visible on the screen
+ */
```
wearhere, line 42: For reference: same comments here re: exceptions as on `SLDatePicker`.

wearhere:
Just cleaning up the documentation now @justinseanmartin, otherwise :ok_hand:

wearhere:
:confetti_ball:! Many thanks @justinseanmartin.

PR #215, opened 2014-06-18
`-[UIView slAccessibilityIsVisible]` reports an element's visibility on-screen, regardless of window.

wearhere:


wearhere:
@aegolden, once Travis finishes, can I get a review on this? I'm particularly interested in
1. whether you think my approach of rendering all the windows into the context is an appropriate way to gauge "screen visibility" (even though my tests pass), and
2. whether I might need to take the windows' sizes into account when creating the rendering buffer. For instance, if a non-key window is larger than the key window (which is what the buffer's size is based on currently), does the buffer need to be made larger?

aegolden:
@wearhere I think this is a great idea.  It makes much more sense for an instance method of UIView to deal with the receiver's window than to take an arbitrary window argument.  The only thing I'm not sure about is whether it's always the case that only windows beyond the receiver's window in allWindows could possibly occlude the view.  What if an app screws around with a window's layer's zPosition?  It might be best to just render all the windows.

wearhere:
I just played around a bit and `zPosition` does not affect window ordering--the equivalent for windows is `windowLevel`, but `-[UIApplication windows]` is sorted by `windowLevel`.

PR #216, opened 2014-06-19
The keyWindow isn't always an application window

justinseanmartin:
This assert was just introduced with PR #211, but it doesn't appear to be valid. We should just handle this condition like the previous behavior.


wearhere:
:bug: :punch:

PR #217, opened 2014-06-19
Allow SLAssert methods to be called from non SLTest subclass methods

justinseanmartin:
What do you think about this? I was planning on building a library of helper methods to build up the test cases, but unless they subclass SLTest, I can't call the assert methods. This doesn't seem right to me. 

This is my proposed solutions, but I wanted to get some thoughts.

### DIFF
--- a/Sources/Classes/SLTest.h
+++ b/Sources/Classes/SLTest.h
@@ -409,10 +409,19 @@
 })
 
 #pragma mark - Test Assertions
+/**
+ The SLAssert* class of methods should only be used from test setup, teardown,
+ or execution methods as well as those methods called from within.
+
+ @warning If you use an assertion inside a dispatch block--if your test case
+ dispatches to the main queue, for instance--you must wrap the assertion in a
+ try-catch block and re-throw the exception it generates (if any) outside the
+ dispatch block. Otherwise, the tests will abort with an unhandled exception.
+ */
 
 /**
  Fails the test case if the specified expression is false.
- 
+
  @param expression The expression to test.
  @param failureDescription A format string specifying the error message 
  to be logged if the test fails. Can be `nil`.
--- a/Sources/Classes/SLTest.h
+++ b/Sources/Classes/SLTest.h
@@ -386,7 +386,7 @@
  @param filename A filename, i.e. the last component of the `__FILE__` macro's expansion.
  @param lineNumber A line number, i.e. the `__LINE__` macro's expansion.
  */
-- (void)recordLastKnownFile:(const char *)filename line:(int)lineNumber;
++ (void)recordLastKnownFile:(const char *)filename line:(int)lineNumber;
 
 /**
  Records the current filename and line number and returns its argument.
@@ -404,7 +404,7 @@
  UIAutomation element corresponding to the wrapped `SLUIAElement`."
  */
 #define UIAElement(slElement) ({ \
-    [self recordLastKnownFile:__FILE__ line:__LINE__]; \
+    [SLTest recordLastKnownFile:__FILE__ line:__LINE__]; \
     slElement; \
 })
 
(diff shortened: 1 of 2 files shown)


wearhere:
A potential concern with letting the assertions be called outside of a test would be that they could be called outside of the current test's exception-handling infrastructure. Would you just leave it up to the user to be aware of/guard against that themselves?

Two alternatives in which you might be interested would be to:
1. define the helpers on the test itself. If the helpers are used by more than one test, you could define them on an [abstract test](http://inkling.github.io/Subliminal/Documentation/Classes/SLTest.html#//api/name/isAbstract). At Inkling, for instance, all tests subclass an "InklingTest" which defines methods related to logging in, downloading books, etc. (Obviously, there's no guarantee that these methods are only called from within test cases either, but I'd say it's more likely.)
2. have helpers defined on other classes return an error/`BOOL` if they succeed or not. You could then have the tests assert that success.

justinseanmartin:
Yea, I'd leave it up to the user to be aware of only throwing these exceptions inside of test code. Let me look at our architecture and see if I can refactor to something else that makes sense.

Sources/Classes/SLTest.m
```diff
@@ -40,10 +40,10 @@
 const NSTimeInterval SLWaitUntilTrueRetryDelay = 0.25;
 
 
-@implementation SLTest {
-    NSString *_lastKnownFilename;
-    int _lastKnownLineNumber;
-}
+@implementation SLTest
+
+static NSString *_lastKnownFilename;
```
wearhere: Please prefix static variables with two underscores to differentiate them from instance variables.

wearhere:
I don't think my objections above are that strong. If you can address my stylistic comment above I'll merge this then.

justinseanmartin:
Done. I also added a comment to SLTest.h on SLAssertTrue for how to do this.

I think there is already a bug saying that we should expand on test case organization, maybe we could add on our use case to that ticket. 

The current design is to break out a class for each screen of our UI with helper methods. This will allow containing of logic for interacting with a particular screen to a single place.

Sources/Classes/SLTest.h
```diff
@@ -404,15 +404,21 @@
...
 #pragma mark - Test Assertions
 
 /**
- Fails the test case if the specified expression is false.
- 
+ Fails the test case if the specified expression is false. The SLAssert class of methods
+ should only be used from test setup, teardown, or execution methods as well as those
+ methods called from within.
+
+ @warning If you perform operations on the main thread within your test case, you can't
```
wearhere: I think this is a good point, though I find the phrasing a bit confusing. What do you think about

> @warning If you use an assertion inside a dispatch block--if your test case dispatches to the main queue, for instance--you must wrap the assertion in a try-catch block and re-throw the exception it generates (if any) outside the dispatch block. Otherwise, the tests will abort.

Sources/Classes/SLTest.h
```diff
@@ -404,15 +404,21 @@
...
 })
 
 #pragma mark - Test Assertions
 
 /**
- Fails the test case if the specified expression is false.
- 
+ Fails the test case if the specified expression is false. The SLAssert class of methods
+ should only be used from test setup, teardown, or execution methods as well as those
+ methods called from within.
```
wearhere: Let's move this sentence and the @warning below into a comment right below the pragma, not tied specifically to `SLAssertTrue`.

wearhere:
Some small comments on documentation. I agree that a discussion of test case organization is out of scope for this request/more appropriate for the wiki.

justinseanmartin:
Incorporated the last 2 pieces of feedback. 

It is unfortunate that the SLAssert documentation doesn't seem to appear in the published documentation anywhere.  It seems that is a limitation of appledoc in general though, so not much you can do about it I guess.

wearhere:
> It is unfortunate that the SLAssert documentation doesn't seem to appear in the published documentation anywhere.

Yeah, I don't think `appledoc` handles macros at present. (The latest release might: #71.)

Looks good! I'll merge as soon as Travis finishes up.

wearhere:
:ok_hand:

PR #220, opened 2014-06-23
Added a hook on SLTest exposing failures that are encountered during a test run.

plantpurecode:
We'd like to display some more debugging information about a test failure when it happens, so we've added a hook to support this.


MaxGabriel:
This seems like a good idea to me. I wonder if Subliminal should also post `NSNotification`s for these kind of things. Subclassing `SLTest` means as soon as you want to add this kind of logging to an existing setup, you need to change the class of all your tests. Also, it means you couldn't make plug-and-play subliminal exception-logger library.

Sources/Classes/SLTest.m
```diff
@@ -232,20 +238,30 @@ + (BOOL)testCaseWithSelectorSupportsCurrentPlatform:(SEL)testCaseSelector {
...
+}
+
+- (void)testRunDidCatchException:(NSException *)exception testCaseSelector:(SEL)testCaseSelector {
+    self.testCaseExceptionInfo = [SLTestCaseExceptionInfo exceptionInfoWithException:exception testCaseSelector:testCaseSelector];
+
+    NSException *exceptionToLog = [self exceptionByAddingFileInfo:self.testCaseExceptionInfo.exception];
+    [[SLLogger sharedLogger] logException:exceptionToLog
+                                 expected:[self.testCaseExceptionInfo isExpected]];
+
+    [self testRunDidCatchExceptionWithExceptionInfo:self.testCaseExceptionInfo];
```
wearhere: I like @MaxGabriel's suggestion to post this information as an `NSNotification`. As Max says that'd be a more flexible arrangement, useful for this purpose and others.

Sources/Classes/SLTestCaseExceptionInfo.h
```diff
@@ -0,0 +1,42 @@
...
+//  Subliminal
+//
+//  Created by Jacob Relkin on 6/20/14.
+//  Copyright (c) 2014 Inkling. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+
+/**
+ `SLTestCaseExceptionInfo` is a class whose objects encapsulate the state of a test run when an exception was caught.
```
wearhere: What do you think about calling this "SLTestCaseFailure"? That's the event that this class models--the failure occurred in this test case, the failure was expected, `SLTest` notifies that a failure did occur, etc.

Sources/Classes/SLTestCaseExceptionInfo.h
```diff
@@ -0,0 +1,42 @@
...
+  */
+
+@interface SLTestCaseExceptionInfo : NSObject
+
+/**
+ Initializes a new `SLTestExceptionInfo` object.
+ @param exception The underlying `NSException` that was thrown. (Required)
+ @param testCaseSelector The test case selector at the time of throw, or NULL if thrown outside of a test case.
+ @see -[SLTest testRunDidCatchExceptionWithExceptionInfo:]
```
wearhere: I think it's a good idea to describe where these models might come from/how they're used, but that info's more appropriate for the class overview than the constructor.

Sources/Classes/SLTest.m
```diff
@@ -265,49 +281,46 @@ - (BOOL)runAndReportNumExecuted:(NSUInteger *)numCasesExecuted
...
-                        caseFailed = YES;
-                        failureWasExpected = [[self class] exceptionWasExpected:exception];
-                        [[SLLogger sharedLogger] logException:[self exceptionByAddingFileInfo:exception]
-                                                     expected:failureWasExpected];
+                        [self testRunDidCatchException:exception testCaseSelector:unfocusedTestCaseSelector];
                     }
                 }
 
+                // If we didn't already fail, testCaseExceptionInfo will be nil
+                BOOL failureWasExpected = self.testCaseExceptionInfo.expected;
```
wearhere: This is a great direction for refactoring `SLTest` and I wonder if we couldn't go even a bit further. What if, rather than tracking the last failure (==exception), `SLTest` had a `testCaseState` property:

``` objc
// just internal to SLTest
@interface SLTestCaseState : NSObject

@property (nonatomic, readonly) BOOL failed;
@property (nonatomic, readonly) BOOL failureWasExpected;

/**
 Sets the properties above, from the failure's description.
 Idempotent, so this can be called multiple times without overwriting `failureWasExpected`. 
 */
- (void)recordFailure:(SLTestCaseFailure *)failure;

@end
```

Then we could get rid of this variable and the tricky ordering in the catch block below.
We also wouldn't have to worry about overwriting `testCaseExceptionInfo` in our exception handler--we actually don't need to hold onto all the exceptions and this would clarify that.

Sources/Classes/SLTest.m
```diff
@@ -265,49 +281,46 @@ - (BOOL)runAndReportNumExecuted:(NSUInteger *)numCasesExecuted
                 // clear call site information, so at the least it won't be reused between test cases
                 // (though we can't guarantee it won't be reused within a test case)
                 [SLTest clearLastKnownCallSite];
+                [self clearTestCaseExceptionState];
```
wearhere: Per [this comment](https://github.com/inkling/Subliminal/pull/220/files#discussion_r14214147), we could just assign `self.testCaseState = [[SLTestCaseState alloc] init];` here.

Sources/Classes/SLTest.m
```diff
@@ -232,20 +238,30 @@ + (BOOL)testCaseWithSelectorSupportsCurrentPlatform:(SEL)testCaseSelector {
     if ([testCaseName hasSuffix:@"_iPad"]) return (userInterfaceIdiom == UIUserInterfaceIdiomPad);
     if ([testCaseName hasSuffix:@"_iPhone"]) return (userInterfaceIdiom == UIUserInterfaceIdiomPhone);
     return YES;
- }
+}
+
+- (void)testRunDidCatchException:(NSException *)exception testCaseSelector:(SEL)testCaseSelector {
```
wearhere: Moving away from the "internal notification" kinda model, I think something action-oriented like "handleException:inTestCase:" might be nicer.

Sources/Classes/SLTest.m
```diff
@@ -232,20 +238,30 @@ + (BOOL)testCaseWithSelectorSupportsCurrentPlatform:(SEL)testCaseSelector {
...
+    self.testCaseExceptionInfo = [SLTestCaseExceptionInfo exceptionInfoWithException:exception testCaseSelector:testCaseSelector];
+
+    NSException *exceptionToLog = [self exceptionByAddingFileInfo:self.testCaseExceptionInfo.exception];
```
wearhere: The failure/exception info ought to be initialized with this exception--it's the most complete record of the event.

Sources/Classes/SLTest.m
```diff
@@ -232,20 +238,30 @@ + (BOOL)testCaseWithSelectorSupportsCurrentPlatform:(SEL)testCaseSelector {
     if ([testCaseName hasSuffix:@"_iPad"]) return (userInterfaceIdiom == UIUserInterfaceIdiomPad);
     if ([testCaseName hasSuffix:@"_iPhone"]) return (userInterfaceIdiom == UIUserInterfaceIdiomPhone);
     return YES;
- }
+}
+
+- (void)testRunDidCatchException:(NSException *)exception testCaseSelector:(SEL)testCaseSelector {
+    self.testCaseExceptionInfo = [SLTestCaseExceptionInfo exceptionInfoWithException:exception testCaseSelector:testCaseSelector];
```
wearhere: Per [this comment](https://github.com/inkling/Subliminal/pull/220/files#discussion_r14214147), this is where we'd record the failure rather than holding onto it.

wearhere:
Thanks for the PR @jacobrelkin! I think this is a great direction, particularly the refactoring of `SLTest`; I've just commented with a couple of ways we could make it cleaner.

plantpurecode:
Addressed all outstanding feedback on this pull request.

@MaxGabriel We decided that posting a notification should not be the concern of `SLTest` itself as the information contained within is bound to it; i.e. the test case selector. To make this data meaningful outside of `SLTest` you'd want to subclass `SLTest` and post a notification from within `-testDidEncounterFailure:`.

Sources/Classes/SLTestCaseExceptionInfo.h
```diff
@@ -0,0 +1,42 @@
+//
+//  SLTestCaseExceptionInfo.h
```
wearhere: It looks like this file and the `.m` aren't used anymore.

Sources/Classes/SLTest.m
```diff
@@ -330,26 +331,38 @@ + (NSString *)unfocusedTestCaseName:(NSString *)testCase {
     return testCase;
 }
 
+- (void)reportFailureInPhase:(SLTestFailurePhase)phase toState:(SLTestState *)state exception:(NSException *)exception testCaseSelector:(SEL)testCaseSelector {
+    SLTestFailure *failure = [SLTestFailure failureWithException:exception phase:phase testCaseSelector:testCaseSelector];
+    NSException *exceptionToLog = [self exceptionByAddingFileInfo:exception];
```
wearhere, line 336: I think that this should be the exception that's recorded in the failure, as the most complete record of the event.

Sources/Classes/SLTest.m
```diff
@@ -330,26 +331,38 @@ + (NSString *)unfocusedTestCaseName:(NSString *)testCase {
     return testCase;
 }
 
+- (void)reportFailureInPhase:(SLTestFailurePhase)phase toState:(SLTestState *)state exception:(NSException *)exception testCaseSelector:(SEL)testCaseSelector {
```
wearhere, line 334: I admire the impulse to keep `SLTest` as stateless as possible though it does make naming this method tricky, haha. 

What do you think about `-reportFailureWithException:inPhase:withTestCaseSelector:state:`? The idea being to keep the test state parameters contiguous. This mirrors `SLTestFailure`'s initializer too.

Sources/Classes/SLTestController.m
```diff
@@ -375,16 +375,20 @@ - (void)runTests:(NSSet *)tests usingSeed:(unsigned int)seed withCompletionBlock
...
-                if (testDidFinish) {
-                    [[SLLogger sharedLogger] logTestFinish:testName
-                                      withNumCasesExecuted:numCasesExecuted
-                                            numCasesFailed:numCasesFailed
-                                numCasesFailedUnexpectedly:numCasesFailedUnexpectedly];
-                    if (numCasesFailed > 0) _numTestsFailed++;
-                } else {
-                    [[SLLogger sharedLogger] logTestAbort:testName];
+
+                [[SLLogger sharedLogger] logTestFinish:testName
```
wearhere: Hm, this looks kinda hairy. It looks like you might be trying to adjust for [changing the return value of `-runAndReport...`](https://github.com/inkling/Subliminal/pull/220/files#diff-4ebb2fdf9f559b7e922291597ddf89cdR446) from concerning just failures in setup/teardown, to any failure?

That would be ok to do if you thought you had a more elegant way, since no one's supposed to be using that method except the test controller, but the contract we've got to maintain is that "finish" is only logged if test set-up and tear-down complete successfully. If a failure occurs in test set-up or tear-down, we must log "abort" instead.

It's kind of an oversight on my part that `-[SLTest testIfTest{Setup,Teardown}FailsAnErrorAndTestAbortAreLogged]` don't reject "finish" being called.

Sources/Classes/SLTestFailure.m
```diff
@@ -0,0 +1,47 @@
...
++ (instancetype)failureWithException:(NSException *)exception phase:(SLTestFailurePhase)phase testCaseSelector:(SEL)testCaseSelector {
+    SLTestFailure *failure = [self new];
+    failure.phase = phase;
+    failure.exception = exception;
+    failure.testCaseSelector = testCaseSelector;
+    return failure;
+}
+
+- (BOOL)isEqual:(id)object {
+    if (![object isKindOfClass:[SLTestFailure class]]) {
```
wearhere, line 31: [`-isEqual:`'s docs](https://developer.apple.com/library/ios/documentation/cocoa/reference/foundation/Protocols/NSObject_Protocol/Reference/NSObject.html#//apple_ref/occ/intfm/NSObject/hash) say that you've got to override `-hash` too.

Unit Tests/SLTestTests.m
```diff
@@ -704,7 +704,8 @@ - (void)testAllTestCasesRunByDefault {
     [[testMock expect] testOne];
     [[testMock expect] testTwo];
     [[testMock expect] testThree];
-    
+    [[testMock reject] testDidEncounterFailure:OCMOCK_ANY];
```
wearhere, line 707: This doesn't seem necessary to have here. (It would actually be possible for the test to encounter a failure and still satisfy the letter of the test case, which is about test-case discovery.)

Unit Tests/SLTestTests.m
```diff
@@ -957,27 +976,37 @@ - (void)runWithTestFailingInTestCaseSetupOrTeardownToTestAnErrorAndTestCaseFailA
     Class failingTestClass = [TestWithSomeTestCases class];
     SEL failingTestCase = @selector(testOne);
     id failingTestMock = [OCMockObject partialMockForClass:failingTestClass];
+    [failingTestMock setExpectationOrderMatters:YES];
```
wearhere, line 979: The expectation sequencer enforces order already, no?

wearhere:
Thanks for all this cleanup @jacobrelkin, removing all those local variables from `-[SLTest runAndReport:...]` is great. I've commented with some remaining functional issues.

I also talked with @justinseanmartin while you (quite understandably) were offline about using notifications after all. I see where you guys are coming from with the test case selectors being test data, but them also representing test cases is a fairly notable concept throughout the framework; and we thought of a couple of nice uses for notifications down the line, where method overrides could get messy. He can fill you in on more of the details.

As a heads-up, I've yet to make a stylistic pass, but I wanted to have you nail the functionality first. Thanks!

Sources/Classes/SLTestController.m
plantpurecode: @wearhere Turns out we don't need this change at all. Reverting.

Sources/Classes/SLTestFailure.m
plantpurecode, line 31: Good call.

Unit Tests/SLTestTests.m
plantpurecode, line 979: Correct. My bad.

PR #226, opened 2014-06-26
Fix environment variable issue for testing on device

hongkai:
When I run the test on device, the environment variables passed to subliminal-test are not available. This patch passed the exact same environment variables to "subliminal-instrument.sh" in launch_instruments() function.

### DIFF
--- a/Supporting Files/CI/subliminal-test
+++ b/Supporting Files/CI/subliminal-test
@@ -158,6 +158,7 @@ SCRIPT_DIR=$(cd "$(dirname "$0")"; pwd)
 BUILD_TOOL="xcrun xcodebuild"
 SCHEME="Integration Tests"
 CONFIGURATION="Release"
+ENV_STR=" "
 
 LOGIN_PASSWORD_DEPRECATION_MESSAGE="""\n
 WARNING: The \`--live\` and \`-login_password\` options have been deprecated\n
@@ -219,6 +220,7 @@ case $1 in
 			elif [[ "$CURRENT_ARG" == "-e" ]]; then
 				# Replace the env var placeholder
 				CURRENT_ARG="export $1"
+				ENV_STR+=" -e $1 $2"
 			else
 				eval $CURRENT_ARG=\"$1\"
 				unset CURRENT_ARG;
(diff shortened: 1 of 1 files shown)


wearhere:
Interesting, I guess that the simulator process is considered a child process of `subliminal-test`, but the device process isn't. Good catch!

With this, we don't need the export logic at all anymore. Also, I don't quite like that the append step looks ahead in the arguments, breaking the argument-by-argument flow of parsing. If the export logic was removed, then `subliminal-test` would fail to warn about environment variables without values (like "-e FOO -e BAR BAZ"). 

Here's a patch that addresses the above points and also clarifies the placeholder logic a bit:

``` diff
diff --git a/Supporting Files/CI/subliminal-test b/Supporting Files/CI/subliminal-test
index f70e9f8..8cf9ea0 100755
--- a/Supporting Files/CI/subliminal-test   
+++ b/Supporting Files/CI/subliminal-test   
@@ -194,7 +194,7 @@ case $1 in
        fi
        case $1 in
            # Set a placeholder for the real arg (the env var)
-           -e) CURRENT_ARG="$1";;
+           -e) CURRENT_ARG="-e";;
            # Variables (apart from env vars) are capitalized in this script
            *)  CURRENT_ARG=`echo ${1#-} | tr [[:lower:]] [[:upper:]]`;;
        esac
@@ -218,9 +218,12 @@ case $1 in
                echo "Value \"$1\" is missing argument"
                print_usage_and_fail
            elif [[ "$CURRENT_ARG" == "-e" ]]; then
-               # Replace the env var placeholder
-               CURRENT_ARG="export $1"
-               ENV_STR+=" -e $1 $2"
+               # Add the variable name to our placeholder
+               CURRENT_ARG="-e $1"
+           elif [[ "$CURRENT_ARG" == "-e "* ]]; then
+               # Add "-e <variable_name> <variable_value>" to the environment string
+               ENV_STR+=" $CURRENT_ARG $1"
+               unset CURRENT_ARG
            else
                eval $CURRENT_ARG=\"$1\"
                unset CURRENT_ARG;
```

If you apply that or something similar, I'll merge. Thanks!

wearhere:
Thanks @hongkai!

hongkai:
You are welcome! Thanks for providing such a powerful framework!

PR #227, opened 2014-07-02
Adding a better description to SLAlertHandler

justinseanmartin:
I thought this would be useful in general when an alert asserts, and wanting more information in the assert itself. We have a scenario where we expect more than 1 alert to be fired and it would be helpful to know which alert isn't being found.


wearhere:
For sure! Thanks.

PR #228, opened 2014-07-06
Fix UIA path for elements within an Alert's view

justinseanmartin:
There is a missing element that appears in the UIA path that is being trimmed in Subliminal. This change is needed in order to find elements in an alert window. I'm not sure if my solution is acceptable of comparing against the class name of _UIModalItemAlertContentView as a string.

The reason for doing this is because we have two alerts that pop up back to back in a particular case. When this happens, we need to ensure that the alerts are handled in a specific order, as the second alert pops in front of the first one while UIA is trying to tap on the first one. This leads to a timing issue and causes the tests to fail randomly. Ignoring the first alert and handling it manually seemed like a simpler solution than trying to modify the alert handler's behavior to respect insertion order.

### DIFF
--- a/Sources/Classes/UIAutomation/User Interface Elements/NSObject+SLAccessibilityHierarchy.m
+++ b/Sources/Classes/UIAutomation/User Interface Elements/NSObject+SLAccessibilityHierarchy.m
@@ -301,7 +301,10 @@
     if ([self.subviews count]) {
         isPopover = [self.subviews[0] isKindOfClass:[UIPopoverBackgroundView class]];
     }
-    return isPopover;
+
+    BOOL isAlertContainer = [NSStringFromClass([self class]) isEqualToString:@"_UIModalItemAlertContentView"];
+
+    return isPopover || isAlertContainer;
 }
 
 // An object is a mock view if its `accessibilityIdentifier` tracks
--- a/Sources/Classes/UIAutomation/User Interface Elements/NSObject+SLAccessibilityHierarchy.m
+++ b/Sources/Classes/UIAutomation/User Interface Elements/NSObject+SLAccessibilityHierarchy.m
@@ -302,9 +302,11 @@
         isPopover = [self.subviews[0] isKindOfClass:[UIPopoverBackgroundView class]];
     }
 
-    BOOL isAlertContainer = [NSStringFromClass([self class]) isEqualToString:@"_UIModalItemAlertContentView"];
+    // Assuming that if the window is not owned by the application, it is probably an alert. Rather than
+    // string checking against a private class, keeping all view's in the path for this case.
+    BOOL isInAlertWindow = ![[[UIApplication sharedApplication] windows] containsObject:[self window]];
 
-    return isPopover || isAlertContainer;
+    return isPopover || isInAlertWindow;
 }
 
 // An object is a mock view if its `accessibilityIdentifier` tracks
--- a/Integration Tests/Tests/SLAlertTest.m
+++ b/Integration Tests/Tests/SLAlertTest.m
@@ -98,6 +98,26 @@
                  @"The handler should have dismissed the alert using the default button.");
 }
 
+- (void)testManuallyHandlingParticularAlertsInTestCode {
+    NSString *cancelButtonTitle = @"Cancel";
+    NSString *defaultButtonTitle = @"Ok";
+
+    // We can mark particular alerts to be left onscreen to be interacted with manually
+    NSString *alertTitle = @"Test Alert";
+    SLAlert *alert = [SLAlert alertWithTitle:alertTitle];
+    SLAlertHandler *handler = [alert dismissByTest];
+    [SLAlertHandler addHandler:handler];
+
+    SLAskApp1(showAlertWithInfo:, (@{   @"title":   alertTitle,
+                                        @"cancel":  cancelButtonTitle,
+                                        @"other":   defaultButtonTitle }));
+    SLAssertTrueWithTimeout([handler didHandleAlert], SLAlertHandlerDidHandleAlertDelay, @"Handler should have handled an alert.");
+    SLAssertNoThrow([[SLButton elementWithAccessibilityLabel:defaultButtonTitle] tap], @"The default button couldn't be found to tap!");
+    SLAssertTrueWithTimeout(SLAskApp(titleOfLastButtonClicked) != nil, SLAlertHandlerDidHandleAlertDelay, @"The alert handler wasn't invoked.");
+    SLAssertTrue([SLAskApp(titleOfLastButtonClicked) isEqualToString:defaultButtonTitle],
+                 @"The handler should have dismissed the alert using the default button.");
+}
+
 - (void)testHandlerMustBeAddedBeforeAlertShows {
     NSString *cancelButtonTitle = @"Cancel";
     NSString *defaultButtonTitle = @"Ok";
--- a/Sources/Classes/UIAutomation/User Interface Elements/SLAlert.h
+++ b/Sources/Classes/UIAutomation/User Interface Elements/SLAlert.h
@@ -106,6 +106,15 @@ typedef NS_ENUM(NSInteger, SLAlertTextFieldType) {
 - (SLAlertDismissHandler *)dismiss;
 
 /**
+ Creates and returns a handler that leaves the matching alert on the screen, and it
+ is up to the test case to find and handle the alert in the view hierarchy to proceed.
+
+ @return A newly created handler that dismisses the corresponding alert using
+ UIAutomation's default procedure.
+ */
+- (SLAlertDismissHandler *)dismissByTest;
+
+/**
  Creates and returns a handler that dismisses a matching alert
  by tapping the button with the specified title.
 
@@ -281,6 +290,8 @@ extern const NSTimeInterval SLAlertHandlerDidHandleAlertDelay;
 @end
 
 
+#if DEBUG
+
 /**
  The methods in the `SLAlert (Debugging)` category are to be used only to 
  debug Subliminal tests.
@@ -305,6 +316,8 @@ extern const NSTimeInterval SLAlertHandlerDidHandleAlertDelay;
 - (SLAlertDismissHandler *)dismissByUser;
 
 @end
+
+#endif
 
 /**
  The methods in the `SLAlertHandler (Debugging)` category may be useful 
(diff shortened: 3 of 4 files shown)


justinseanmartin:
Also, I know I should add a test case to validate this scenario continues to work as expected (hence WIP in the name).

wearhere:
Thanks for putting up the WIP @justinseanmartin. If I understand your problem, it goes like this:
1. you register handlers for alerts A and B, in that order
2. A shows
3. A's handler begins
4. _before A's handler finishes_, B shows, displacing A
5. B's handler does _not_ run, because A's hasn't finished--and A's handler can't finish, because A is no longer visible.

Is that right? In particular, can you confirm that B's handler does not run? If that's right, I'm not sure how Subliminal could detect this…

By the way, the reason why Subliminal recommends against manipulating alerts directly is because system alerts can't be so handled--they're remote views. But in this case I guess you've got to, and are able to.

To truly resolve timing issues, I would _prevent_ Subliminal from attempting to handle alert A by registering a handler like `dismissByUser` (vs. assuming that Subliminal's handler will just fail). However, that handler's only available `if #DEBUG`. We could either remove that conditional, or introduce a similar handler `dismissByTest` with some documentation more apropos to this situation, maybe in a public-but-not-included-in-`Subliminal.h` header. What do you think?

Sources/Classes/UIAutomation/User Interface Elements/NSObject+SLAccessibilityHierarchy.m
```diff
@@ -301,7 +301,10 @@ - (BOOL)classForcesPresenceInAccessibilityHierarchy {
     if ([self.subviews count]) {
         isPopover = [self.subviews[0] isKindOfClass:[UIPopoverBackgroundView class]];
     }
-    return isPopover;
+
+    BOOL isAlertContainer = [NSStringFromClass([self class]) isEqualToString:@"_UIModalItemAlertContentView"];
```
wearhere: I'd like to continue to avoid referencing private symbols if possible. Any chance that you can identify this view by its position in the hierarchy, e.g. as the nth child of `UIAlertView`?

justinseanmartin:
You got it mostly right. While the alert handler for Alert A is running, the view object backing it can become untappable (displaced by Alert B on the screen). It isn't that the handler for Alert B doesn't run, but the handler for Alert A can't complete by tapping a button on Alert A when Alert B is in front of it, generating a failure.

I had to work around this exact same issue in my direct UIAutomation scripts previously. The way I was handling it was to never interact with an alert directly within the alert handler, but instead always just return true. It would then be up to each test case to know when alerts are expected and look for the UIAAlert objects in the view hierarchy when appropriate. In the case of this test, the test case knew to wait for "Alert B" to be visible before interacting with it to dismiss, and only then handling "Alert A". One of my testSetup steps was to clear any visible alerts that may have been leftover from a previous test to ensure the app was in a good state.

Outside of the scope of this PR, I'd eventually like to get back into a state where I can:
1. Validate alert message text as well as title
2. Validate alert button text (without needing to click on both of them)
3. Generate a warning/failure on unexpected alerts appearing

Sources/Classes/UIAutomation/User Interface Elements/NSObject+SLAccessibilityHierarchy.m
justinseanmartin: Is the issue that this will be caught during static analysis for app store submission? I'll see what I can do, but I don't think I saw a UIAlertView or anything else really helpful in the view hierarchy. I'll get the output tomorrow.

wearhere: If a user linked Subliminal against their main target rather than (or in addition to) a testing target, then yes this would raise a flag during app store submission. I don't expect users to link Subliminal against their main target, but it's really nice to not have to tell them not to like, say, [KIF does](https://github.com/kif-framework/KIF#kif-ios-integration-testing-framework). It's also a big marketing win to say that we _don't_ use private APIs.

justinseanmartin: I just posted a another potential fix that doesn't require the reference to the private class. The minimum UIA path would be:

Target -> Alert Window -> Alert View -> Table View -> Table Cell

The "Alert View" is really the _UIModalItemAlertContentView and is the one required element not making it into the accessibility path. However, leaving all UIView's in the path in this case seems to work fine as well. I'm testing this on iOS 6 now to make sure it works in general.

This is what the alert window hierarchy contains:

> <_UIModalItemHostingWindow: 0x14815d70; frame = (0 0; 320 568)>
>   | <UIView: 0x14815a60; frame = (0 0; 320 568)>
>   | <UIView: 0x18336af0; frame = (0 0; 320 568)>
>   |   | <UIView: 0x183573b0; frame = (0 0; 320 568)>
>   |   | <_UIModalItemRepresentationView: 0x183565c0; frame = (25 212.5; 270 143)>
>   |   |   | <_UIModalItemAlertBackgroundView: 0x1832cfb0; frame = (25 212.5; 270 143)>
>   |   |   |   | <_UIBackdropView: 0x183487d0; frame = (25 212.5; 270 143)>
>   |   |   |   |   | <_UIBackdropEffectView: 0x18322b10; frame = (25 212.5; 270 143)>
>   |   |   |   |   | <UIView: 0x183383b0; frame = (25 212.5; 270 143)>
>   |   |   |   | <UIImageView: 0x183455a0; frame = (25 212.5; 270 143); traits = (Image)>
>   |   |   | <_UIModalItemAlertContentView: 0x148150a0; frame = (25 212.5; 270 143)>
>   |   |   |   | <UITableView: 0x11a94c00; frame = (25 311.5; 270 44); label = 'Empty list'>
>   |   |   |   |   + <UITableViewCellAccessibilityElement: 0x1835b470; frame = (25 311.5; 270 44); label = 'OK'; traits = (Button); accessibilityElement = YES>
>   |   |   |   |   +   + <UITableTextAccessibilityElement: 0x18325260; frame = (25 311.5; 270 44); label = 'OK'; traits = (Button, Static Text); accessibilityElement = YES>
>   |   |   |   |   | <UITableViewWrapperView: 0x1834cb20; frame = (25 311.5; 270 44)>
>   |   |   |   |   |   | <_UIModalItemTableViewCell: 0x183504c0; frame = (25 311.5; 270 44); label = 'OK'; traits = (Button); accessibilityElement = YES>
>   |   |   |   |   |   |   + <UITableTextAccessibilityElement: 0x18325260; frame = (25 311.5; 270 44); label = 'OK'; traits = (Button, Static Text); accessibilityElement = YES>
>   |   |   |   |   |   |   | <UITableViewCellScrollView: 0x18357080; frame = (25 311.5; 270 44)>
>   |   |   |   |   |   |   |   | <UITableViewCellContentView: 0x18345ef0; frame = (25 311.5; 270 43.5)>
>   |   |   |   |   |   |   |   |   | <UILabel: 0x1834a8c0; frame = (40 311.5; 240 43.5); label = 'OK'; traits = (Static Text); accessibilityElement = YES>
>   |   |   |   |   |   |   |   |   | <UIImageView: 0x10ef95c0; frame = (25 311.5; 0 0); traits = (Image)>
>   |   |   |   |   | <UIImageView: 0x18357800; frame = (25 352; 270 3.5); traits = (Image)>
>   |   |   |   |   | <UIImageView: 0x18357a50; frame = (291.5 311.5; 3.5 44); traits = (Image)>
>   |   |   |   | <UIScrollView: 0x1834cf10; frame = (25 212.5; 270 80)>
>   |   |   |   |   | <UILabel: 0x1834beb0; frame = (40 231.5; 240 21); label = 'Error'; traits = (Static Text); accessibilityElement = YES>
>   |   |   |   |   | <UILabel: 0x1834c310; frame = (40 256.5; 240 36); label = 'That email is already taken. Please enter another address.'; traits = (Static Text); accessibilityElement = YES>
>   |   |   |   |   | <UIImageView: 0x18358fb0; frame = (25 289; 270 3.5); traits = (Image)>
>   |   |   |   |   | <UIImageView: 0x183591e0; frame = (291.5 212.5; 3.5 80); traits = (Image)>
>   |   |   |   | <UIView: 0x18355100; frame = (25 311; 270 0.5)>

PS - Web formatting is hard.

Sources/Classes/UIAutomation/User Interface Elements/NSObject+SLAccessibilityHierarchy.m
```diff
@@ -301,7 +301,12 @@ - (BOOL)classForcesPresenceInAccessibilityHierarchy {
     if ([self.subviews count]) {
         isPopover = [self.subviews[0] isKindOfClass:[UIPopoverBackgroundView class]];
     }
-    return isPopover;
+
+    // Assuming that if the window is not owned by the application, it is probably an alert. Rather than
+    // string checking against a private class, keeping all view's in the path for this case.
+    BOOL isInAlertWindow = ![[[UIApplication sharedApplication] windows] containsObject:[self window]];
```
wearhere, line 307: This new fix looks great. This will probably only catch system views--not sure why a window wouldn't be vended by `UIApplication` otherwise--and might even help us catch other system views.

wearhere:
I think I'd prefer adding a `dismissByTest` handler to using `dismissByUser` for this purpose (though they'd both do the same thing)--the former will make tests clearer.

You can put `dismissByTest` in the main header, though--In light of the potential for this issue, I think I'd actually like to amend `SLAlert`'s docs to recommend direct manipulation if possible. (Doesn't hurt that it's easier too.) Modifying the `SLAlertHandler` to always return `true` would be pretty sensible also if you could ensure that unhanded alerts would fail test cases rather than just hang. These changes are probably out of scope for this request, though.

Once you've fixed up `dismissByTest` and added a test for this--you can just mimic an `SLAlert` case, except dismiss the alert directly--this will be good to merge.

justinseanmartin:
Added a separate dissmissByTest handler and added a test case. For the more radical replacing the default handler to always return true, I'll submit a separate PR for that at a later time, but I think that approach is much more straightforward.

justinseanmartin:
Test that failed was the timing issue "testWaitUntilTrueReturnsNOIfConditionIsStillFalseAtEndOfTimeout"

wearhere:
Dang it Travis... https://travis-ci.org/inkling/Subliminal/jobs/29647521 taking 14 minutes is pretty weird too. I'm going to ignore that for the meantime--these changes look good. Thanks!

PR #229, opened 2014-07-08
Wait for elements to be valid before attempting to type in them

justinseanmartin:
Wait for elements to be valid before attempting to type in them. I was finding that sometimes the "type" would be failing because an alert window was still the keyWindow when [self hasKeyboardFocus] would attempt to generate an accessibilityPath.


Sources/Classes/UIAutomation/User Interface Elements/SLTextField.m
```diff
@@ -35,6 +35,9 @@ - (void)setText:(NSString *)text {
 
 - (void)setText:(NSString *)text withKeyboard:(id<SLKeyboard>)keyboard
 {
+    // Need to make sure the element is visible first, because hasKeyboardFocus is a one-shot check
+    [self waitUntilValidWithTimeout:[SLElement defaultTimeout]];
+
     // Tap to show the keyboard (if the field doesn't already have keyboard focus,
     // because in that case a real user would probably not tap again before typing)
     if (![self hasKeyboardFocus]) {
```
wearhere, line 43: I think it would be simpler just to amend this line to read `if (![self isValid] || ![self hasKeyboardFocus]) {…`. Then we can use `tap` to wait for validity.

wearhere:
Good catch, @justinseanmartin--`-setText` is supposed to wait till the element is valid. Can you fix this in `SLWebTextField` and `SLTextView/SLWebTextView` too?

wearhere:
:+1: Thanks!

PR #230, opened 2014-07-09
When an alert handler fails to run, log useful information

justinseanmartin:
When an alert handler fails to run, log some useful information. When I added this, it appears subliminal-instrument would crash with an index out of bounds exception.

Thoughts?

### DIFF
--- a/Sources/Classes/UIAutomation/User Interface Elements/SLAlert.m
+++ b/Sources/Classes/UIAutomation/User Interface Elements/SLAlert.m
@@ -135,15 +135,21 @@ static BOOL SLAlertHandlerLoggingEnabled = NO;
                 // enumerate registered handlers, from first to last
                 @"for (var handlerIndex = 0; handlerIndex < SLAlertHandler.alertHandlers.length; handlerIndex++) {\
                     var handler = SLAlertHandler.alertHandlers[handlerIndex];"
-                    // if a handler matches the alert...
-                    @"if (handler.handleAlert(alert) === true) {\
-                        if (SLAlertHandler.loggingEnabled) UIALogger.logMessage('Alert was handled by a test.');"
-                        // ...ensure that the alert's delegate will receive its callbacks
-                        // before the next JS command (i.e. -didHandleAlert) evaluates...
-                        @"UIATarget.localTarget().delay(%g);"
-                        // ...then remove the handler and return true
-                        @"SLAlertHandler.alertHandlers.splice(handlerIndex, 1);\
-                        return true;\
+                    @"try {"
+                        // if a handler matches the alert...
+                        @"if (handler.handleAlert(alert) === true) {\
+                            if (SLAlertHandler.loggingEnabled) UIALogger.logMessage('Alert was handled by a test.');"
+                            // ...ensure that the alert's delegate will receive its callbacks
+                            // before the next JS command (i.e. -didHandleAlert) evaluates...
+                            @"UIATarget.localTarget().delay(%g);"
+                            // ...then remove the handler and return true
+                            @"SLAlertHandler.alertHandlers.splice(handlerIndex, 1);\
+                            return true;\
+                        }\
+                    } catch (e) {\
+                        UIALogger.logError('Alert handler threw an exception!');\
+                        UIATarget.localTarget().logElementTree();\
+                        throw e;\
                     }\
                 }"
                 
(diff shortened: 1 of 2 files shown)


Supporting Files/CI/subliminal-instrument/subliminal-instrument/SISLLogParser.m
```diff
@@ -157,8 +157,10 @@ + (void)parseEventType:(SISLLogEventType *)type subtype:(SISLLogEventSubtype *)s
 
                     // message format: "SLKeyboardTest.m:62: ..."
                     NSArray *messageComponents = [message componentsSeparatedByString:@":"];
-                    infoValue[@"fileName"] = messageComponents[0];
-                    infoValue[@"lineNumber"] = @([messageComponents[1] integerValue]);
+                    if (messageComponents.count > 2) {
```
wearhere, line 162: Ah good catch. I guess some of `instruments`' exceptions begin with "Error:" too. Can you amend the comment at 158 to say as much?

wearhere, line 162: Actually I think a more apt description would be: "In rare circumstances, Subliminal may log an error with no call site." (the error's not coming from `instruments`--it's our use of `UIALogger.logError` above).

Sources/Classes/UIAutomation/User Interface Elements/SLAlert.m
```diff
@@ -135,15 +135,21 @@ + (void)loadUIAAlertHandling {
...
+                            if (SLAlertHandler.loggingEnabled) UIALogger.logMessage('Alert was handled by a test.');"
+                            // ...ensure that the alert's delegate will receive its callbacks
+                            // before the next JS command (i.e. -didHandleAlert) evaluates...
+                            @"UIATarget.localTarget().delay(%g);"
+                            // ...then remove the handler and return true
+                            @"SLAlertHandler.alertHandlers.splice(handlerIndex, 1);\
+                            return true;\
+                        }\
+                    } catch (e) {\
+                        UIALogger.logError('Alert handler threw an exception!');\
```
wearhere: How about `"Handler for alert " + alert.name() +`" threw an exception!` (If logging's not enabled, the name won't be printed above this.)

Sources/Classes/UIAutomation/User Interface Elements/SLAlert.m
```diff
@@ -135,15 +135,21 @@ + (void)loadUIAAlertHandling {
...
+                        UIATarget.localTarget().logElementTree();\
+                        throw e;\
```
wearhere: Are you sure that it's ok to throw from inside `onAlert`? I could have sworn that doing so crashed `instruments`, at least the GUI.

wearhere:
A few notes, back to you @justinseanmartin.

Sources/Classes/UIAutomation/User Interface Elements/SLAlert.m
justinseanmartin: It does, I was just going to let that happen... I could not rethrow, in which case it should fall through to the default handler. I suppose that is preferable? Would you still splice out the handler in this case?

wearhere: Yeah, I think we might as well try to let the default handler get rid of the alert rather than abort the run (like `onAlert` does currently if the alert doesn't look like it can be dismissed at all). If it can't then the run's effectively over anyway.

I'd still splice out the handler, yeah. Each handler only gets one shot at dismissing a matching alert and this handler tried for this one.

justinseanmartin:
All comments have been addressed, and I also added a 2 second UIA timeout around interacting with Alert buttons.

Sources/Classes/UIAutomation/User Interface Elements/SLAlert.m
```diff
@@ -393,7 +399,12 @@ - (SLAlertDismissHandler *)dismissWithButtonTitled:(NSString *)buttonTitle {
     NSString *UIAAlertHandler = [NSString stringWithFormat:@"\
                                      var buttonElement = (%@)('%@');\
                                      if (buttonElement && buttonElement.isValid()) {\
-                                        buttonElement.tap();\
+                                        UIATarget.localTarget().pushTimeout(2);\
```
wearhere: What do you think about the `onAlert` function pushing/popping the timeout? Then that would cover any handler, not just `-dismissWithButtonTitled:`.

wearhere: Also let's make the timeout 5 seconds--that's the default (used by UIAutomation absent Subliminal; and by all of Subliminal's APIs).

justinseanmartin: The problem with this is that we're going to wait 5 seconds for each registered alert that doesn't match on the title. I think I can fix it by modifying the dismissWithButton logic though.

justinseanmartin: I'm wrong, thanks for correcting me.

wearhere: No problem, thanks for double-checking!

wearhere:
Looks good, thanks!

PR #231, opened 2014-07-10
Try and handle NSData UTF8 encoding failures in subliminal-instruments

justinseanmartin:
The issue with the current logic is that it assumes that each individual data buffer will successfully convert into a UTF8 string. However, this is not the case if instruments flushes the pipe in the middle of a UTF8 multi-byte character, and will lead to an exception when we try to appendString:nil to buffer.

The easy solution to this crash is that if the data buffer doesn't convert, then wait for the next data buffer. The assumption is that before the pipe is closed, we'll receive a buffer that can be successfully converted.

FYI - One potential issue with this solution is that if the Instruments output has a lot of non-ASCII text in it, the data buffer will be more likely to fail conversion (and grow the buffer in memory while doing so). This is could possibly happen with a non-English localized app-under-test (e.g. Japanese). This could be even more problematic if the OSX system language were set to a different language.

### DIFF
--- a/Supporting Files/CI/subliminal-instrument/subliminal-instrument/NSTask+Utilities.m
+++ b/Supporting Files/CI/subliminal-instrument/subliminal-instrument/NSTask+Utilities.m
@@ -47,25 +47,32 @@ void SIRegisterHandlerForLinesReadFromHandle(NSFileHandle *handle, void(^lineHan
 
     // create a buffer to store data until a full line is read
     // the readability handler will retain the buffer
-    NSMutableString *buffer = [[NSMutableString alloc] init];
+    NSMutableData *charBuffer = [[NSMutableData alloc] init];
+    __block NSUInteger charOffset = 0;
 
     [handle setReadabilityHandler:^(NSFileHandle *handle) {
         @autoreleasepool {
-            NSString *currentText = [[NSString alloc] initWithData:[handle availableData] encoding:NSUTF8StringEncoding];
-            [buffer appendString:currentText];
-
-            NSScanner *newlineScanner = [[NSScanner alloc] initWithString:buffer];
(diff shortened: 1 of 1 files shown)


justinseanmartin:
Credit for the fix goes to @wearhere. He cleaned up my quick and dirty fix and instead made the right fix. The new branch commit no longer should exhibit the issues that I called out as the potential issues.

Supporting Files/CI/subliminal-instrument/subliminal-instrument/NSTask+Utilities.m
```diff
@@ -47,25 +47,32 @@ void SIRegisterHandlerForLinesReadFromHandle(NSFileHandle *handle, void(^lineHan
...
+            [charBuffer appendData:[handle availableData]];
+            
+            // We manually search for newline characters rather than, say, encoding the string
+            // and using an `NSScanner`, because we may not be able to encode the current buffer
+            // if we've read less than a full UTF-8 multibyte character
+            char *chars = (char *)[charBuffer bytes];
+            NSUInteger lineStart = 0;
+            for (; charOffset < [charBuffer length] / sizeof(char); charOffset++) {
+                if (chars[charOffset] == '\n') {
+                    NSData *lineData = [charBuffer subdataWithRange:NSMakeRange(lineStart, (charOffset - 1) * sizeof(char))];
```
wearhere: I messed this up per Travis. I think it should be `charOffset - 1 - lineStart`?

justinseanmartin: Fixed up, running locally and on CI now, so far so good.

wearhere:
Thanks @justinseanmartin for exploring the edge cases here--and, I must note, suggesting the idea that I implemented, and then fixing my bug in that implementation. ;) Cheers.

PR #232, opened 2014-07-10
Visibility matching on SLElement

plantpurecode:
Added functionality to SLElement which further filters the matching routine based on the visibility of a view.

### DIFF
--- a/Sources/Classes/UIAutomation/User Interface Elements/SLElement.m
+++ b/Sources/Classes/UIAutomation/User Interface Elements/SLElement.m
@@ -135,6 +135,13 @@ UIAccessibilityTraits SLUIAccessibilityTraitAny = 0;
     NSAssert(_matchesObject, @"matchesObject called on %@, which has no _matchesObject predicate", self);
     BOOL matchesObject = _matchesObject(object);
 
+    // Non-UIView objects are assumed to be "not hidden"
+    if (self.matchOnlyIfVisible) {
+        if ([object respondsToSelector:@selector(isHidden)]) {
+            matchesObject = matchesObject && ![(UIView *)object isHidden];
+        }
+    }
+
     return (matchesObject && [object willAppearInAccessibilityHierarchy]);
 }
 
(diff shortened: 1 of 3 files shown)


Sources/Classes/UIAutomation/User Interface Elements/SLElement.m
```diff
@@ -135,6 +135,13 @@ - (BOOL)matchesObject:(NSObject *)object
     NSAssert(_matchesObject, @"matchesObject called on %@, which has no _matchesObject predicate", self);
     BOOL matchesObject = _matchesObject(object);
 
+    // Non-UIView objects are assumed to be "not hidden"
+    if (self.matchOnlyIfVisible) {
```
wearhere, line 139: You'd want to use [`-[NSObject slAccessibilityIsVisible]`](https://github.com/inkling/Subliminal/blob/master/Sources/Classes/Internal/NSObject+SLVisibility.h#L40), and check all objects, here--there are a bunch of reasons why objects, including non-views, might not be visible from both the user's and UIAutomation's perspectives.

(Making this change would cause `matchOnlyIfVisible` to fail in non-portrait orientations (https://github.com/inkling/Subliminal/issues/135), but would be generally more robust. It would be nice if you could fall back to `UIAElement.isVisible` as does [`-[SLElement isVisible]`](https://github.com/inkling/Subliminal/blob/master/Sources/Classes/UIAutomation/User%20Interface%20Elements/SLElement.m#L346) but I'm not sure how you'd do that given that matching takes place on the main thread.)

wearhere:
Is the scenario here that you have two elements in the hierarchy with the same attributes, only one is visible and one isn't? How often would you say this comes up? An alternative to building this functionality into `SLElement` would be to use `-elementMatching:` as shown [here](https://github.com/inkling/Subliminal/wiki/Debugging-Tests#why-does-subliminal-say-that-an-slelement-doesnt-existisnt-visibleisnt-tappable-when-i-can-seetap-it-just-fine).

plantpurecode:
Will use `-[NSObject slAccessibilityIsVisible]`, thanks!

PR #233, opened 2014-07-16
Add support for tagging tests and test cases.

wearhere:
Delivers the following features:
- tests and test cases can be tagged, with sensible defaults (test/test case names)
- `+[SLTest testsWithTags:]` allows a subset of tests to be selected while debugging
- the `SL_TAGS` environment variable allows a subset of tests and/or test cases to be selected in CI
- tests/test cases can be excluded by prefixing tags with "-"
- tests and test cases can respond to additional environment variables by overriding `+supportsCurrentEnvironment` and `+testCaseWithSelectorSupportsCurrentEnvironment:`

Many thanks to all (@jazzychad, @qiffp, @maxgabriel, @mrsamuelkim) who have weighed in on this topic. I think that this solution is general purpose enough to encompass the majority of your use cases.

In particular, it allows particular tests _or_ groups of tests to be selected, because tests are automatically tagged with their names; it does not interfere with the focus mechanism; it does not require that tagging is exhaustive, because tags can be excluded; and not only tests but also test cases can be tagged, using the same mechanism.

### DIFF
--- a/Sources/Classes/SLTest.h
+++ b/Sources/Classes/SLTest.h
@@ -47,7 +47,7 @@
  
  Without modifying the argument to `-[SLTestController runTests:withCompletionBlock:]`, 
  tests may be conditionalized to run only in certain circumstances using APIs
- like `+isAbstract`, `+supportsCurrentPlatform`, and `+isFocused`.
+ like `+isAbstract`, `+supportsCurrentPlatform`, `+supportsCurrentEnvironment`, and `+isFocused`.
 
  @return All tests (`SLTest` subclasses) linked against the current target.
  */
@@ -88,13 +88,13 @@
 + (BOOL)isAbstract;
 
 /**
- Returns YES if this test has at least one test case which can be run
+ Returns YES if this test has at least one test case which should be run
  given the current device, screen, etc.
  
  Subclasses of `SLTest` should override this method if some run-time condition
- should determine whether or not all test cases should run. 
- Typical checks might include checking the user interface idiom (phone or pad) 
- of the current device, or checking the scale of the main screen.
+ concerning the current platform should determine whether or not all test cases
+ should be run. Typical checks might include checking the user interface idiom
+ (phone or pad) of the current device, or checking the scale of the main screen.
 
  As a convenience, test writers may specify the device type(s) on which a
  test can run by suffixing tests' names in the following fashion:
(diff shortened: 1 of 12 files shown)


Sources/Classes/SLTest.h
```diff
@@ -111,17 +177,134 @@
...
+ ("FooTest", "1", and "foo") and the following values of `SL_TAGS`,
+ this method would return:
+ 
+ * (not set) -> `YES`
+ * "FooTest" -> `YES` (because `-foo` is tagged with "FooTest")
+ * "foo" -> `YES` (because `-foo` is tagged with "foo")
+ * "-bar" -> `YES` (because `-foo` is not tagged with "bar")
+ * "-foo" -> `NO` (because `-foo` is tagged with "foo")
+ * "bar" -> `NO` (because `-foo` is not tagged with "bar")
+ * "-bar,baz" -> `NO` (because `-foo` is not tagged with "baz")
```
justinseanmartin, line 287: The same (all conditions must be true) is also holds for the opposite case, correct? E.g. "FooTest,-foo".

Sources/Classes/SLTest.m
```diff
@@ -110,7 +154,60 @@ + (BOOL)supportsCurrentPlatform {
         testClass = [testClass superclass];
     }
 
-    return testSupportsCurrentDevice;
+    BOOL aTestCaseSupportsCurrentPlatform = NO;
+    for (NSString *testCaseName in [self testCases]) {
```
justinseanmartin, line 158: NIT: You only really need to do this if testSupportsCurrentDevice is true. The way it is now might be easier to read even if it is less efficient than having an early return or wrapping it in an if (testSupportsCurrentDevice).

Sources/Classes/SLTest.h
wearhere, line 287: If `SL_TAGS` were set to "FooTest,-foo", `[FooTest testCaseWithSelectorSupportsCurrentEnvironment:@selector(foo)]` would return `NO`, because `foo` is tagged with "foo". ("FooTest" includes it, but "foo" excludes it.) So I think the answer to your question is "yes". Is that what you expected?

I think I'll add this case to the list too.

Sources/Classes/SLTest.m
wearhere, line 158: Yeah, I chose to do it this way for readability. I also don't expect this to have to enumerate all test cases in any but the rarest situations. Good thought though.

Sources/Classes/SLTest.h
justinseanmartin, line 287: Yep, that's what I'd expect.

justinseanmartin:
LGTM - I got a bit lazy in reviewing the unit tests in detail, but design looks good to me and the high level of the unit tests looked good. Thanks for putting this together, it will definitely be useful!

ghost:
Looks good to me, @wearhere! Thank you for your work. It will definitely be useful for us.

qiffp:
Looks good to me as well. Thanks for putting this together, @wearhere!

wearhere:
Thanks for looking it over, everyone! @aegolden / @ahaneyinkling, can one of you review and merge, please?

MaxGabriel:
Its not affecting me personally, but this seems like a major feature for several users. Given that + all of @justinseanmartin's PRs lately, do you think its time for another release?

wearhere:
Good call! I've updated the CHANGELOG to cover the work done since v1.1.0, and pulled out the remaining issues I'd like to tackle for 1.2 [here](https://github.com/inkling/Subliminal/issues?milestone=3&page=1&state=open).

This will be 1.2. not 2.0 like we had planned, because I don't want to release 2.0 until iOS 8 is released, because we'll obviously have to release a version then to support iOS 8, but we'll _also_ have to drop support for iOS 6 at that time. (Xcode 6 doesn't support running iOS 6 even if you copy the SDK into its developer directory.)

That is, if we went 2.0 now, we'd have to go _3.0_ when iOS 8 was released, and I know semver says that's fine but it seems a little silly to me to jump 2 major versions within the span of two months. :)

What makes _not_ going 2.0 just yet ok is that I found that Inkling still has an iPod touch and an iPad both running iOS 5.1, which means I can test Subliminal against iOS 5 and not drop compatibility with it as we had feared would be necessary after the upgrade of Travis (and our local machines) to Mavericks.

PR #249, opened 2014-08-05
Added an integration test for UITextField in UICollectionViewCell

jzucker2:
Found some issues we had with UICollectionViewCells and a child
UITextField. Created an integration test for regression checking.


jzucker2:
Just to give a little context. Was having some troubles with UITextFields in a UICollectionViewCell. Wanted to prove that it is a problem in my UI implementation and not an issue with Subliminal itself. Do you think this is a good test?

PR #251, opened 2014-08-08
Add a class property for the default value for SLElement's shouldDoubleCheckValidity

justinseanmartin:
We'd like to turn on shouldDoubleCheckValidity all the time, and this seemed like the most straightforward way to expose this functionality. Turning it on has helped with automation reliability a bit. I honestly don't see an issue with having this running all the time, but that's your call.

### DIFF
--- a/Sources/Classes/UIAutomation/User Interface Elements/SLElement.h
+++ b/Sources/Classes/UIAutomation/User Interface Elements/SLElement.h
@@ -207,4 +207,11 @@ extern UIAccessibilityTraits SLUIAccessibilityTraitAny;
  */
 @property (nonatomic) BOOL shouldDoubleCheckValidity;
 
+/**
+ The shouldDoubleCheckValidity BOOL defaults to NO, but this allows making double checking for the existence
+ of elements the default behavior for newly created SLElement objects, essentially making it opt-out instead
+ of opt-in.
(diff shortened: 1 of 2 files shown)


wearhere:
This is a great find, @justinseanmartin. If you can clearly see that this resolves an issue you're having, I'm fine not only taking this but making it the default. Worst case scenario, it can cause about a half second delay per match for the cost of the round trip evaluation of the command (see `SLElementTapTest` for such a calculation), but I'd rather optimize for safety than performance. People could toggle it off if they wanted.

I would like to be clear about what we're doing though, which isn't really changing the default state of a diagnostic. Double-checking is merely a means to... at a guess, warming and/or purging a cache within UIAutomation in advance of actually manipulating the interface. Can we add a new flag like `shouldPrefetchElements` to describe it as such? (Double-checking would then be controlled by `_shouldDoubleCheckValidity || __shouldPrefetchElements`).

wearhere:
@justinseanmartin IIRC you said in chat that this didn't actually have the effect you thought? If so can you close?

justinseanmartin:
It did make some differences, but it required other changes as well, such as expanding the types of exceptions that would be retried as well. Instead, we opted to disable animations in the app which has helped with test stability.

PR #253, opened 2014-08-15
Add tests' superclasses' names to the default tag set. (fixes #252)

### LINKED ISSUE #252: Class tag NSSet could also include superclasses (up to SLTest)
Just a thought, but in our case, we are tagging our classes much in the same way that we're organizing our class hierarchies. It might be nice if SLTest derivative classes also tagged themselves with their superclasses.

wearhere:


justinseanmartin:
LGTM!

wearhere:
Thanks for the review @justinseanmartin!

PR #255, opened 2014-08-26
Apply focus before filtering for platform/environment support.

wearhere:
Focus is no longer ignored in cases where focused tests wouldn't run--now, no tests run. This makes it easier to reason about what will happen when you focus a test: no unfocused tests will run.

### DIFF
--- a/Sources/Classes/SLTestController.m
+++ b/Sources/Classes/SLTestController.m
@@ -349,6 +349,9 @@ u_int32_t random_uniform(u_int32_t upperBound) {
 }
 
 - (void)runTests:(NSSet *)tests usingSeed:(unsigned int)seed withCompletionBlock:(void (^)())completionBlock {
+    // have to check this outside of the block below, wherein it will be used
+    static const char *const kMethodDescription = __PRETTY_FUNCTION__;
+    
     dispatch_async(_runQueue, ^{
         _completionBlock = completionBlock;
 
@@ -356,7 +359,14 @@ u_int32_t random_uniform(u_int32_t upperBound) {
         _runSeed = seed;
         _testsToRun = [[self class] testsToRun:tests usingSeed:&_runSeed withFocus:&_runningWithFocus];
         if (![_testsToRun count]) {
-            SLLog(@"%@%@%@", @"There are no tests to run", (_runningWithFocus) ? @": no tests are focused" : @"", @".");
+            NSMutableString *noTestsToRunWarning = [@"There are no tests to run: " mutableCopy];
+            if ([tests count]) {
+                [noTestsToRunWarning appendFormat:@"none of the tests %@ meet the criteria to be run. See `%@`'s documentation.",
+                                                 (_runningWithFocus) ? @"focused" : @"passed", @(kMethodDescription)];
+            } else {
+                [noTestsToRunWarning appendString:@"no tests were passed."];
+            }
+            [[SLLogger sharedLogger] logWarning:noTestsToRunWarning];
             [self _finishTesting];
             return;
         }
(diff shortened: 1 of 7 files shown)


Sources/Classes/SLTestController.m
```diff
@@ -191,16 +191,7 @@ + (NSArray *)testsToRun:(NSSet *)tests usingSeed:(inout unsigned int *)seed with
     }
 
     // now filter the tests to run:
```
justinseanmartin: SUPER NIT: The comment looks strange like this

wearhere: It's to try to break the comment over multiple lines/interleave it with the code, you can see with the ellipses below `*hangs head sheepishly*`... suggestions for how I might improve it?

Sources/Classes/SLTestController.m
```diff
@@ -356,7 +359,15 @@ - (void)runTests:(NSSet *)tests usingSeed:(unsigned int)seed withCompletionBlock
...
         if (![_testsToRun count]) {
-            SLLog(@"%@%@%@", @"There are no tests to run", (_runningWithFocus) ? @": no tests are focused" : @"", @".");
+            NSMutableString *noTestsToRunWarning = [@"There are no tests to run: " mutableCopy];
+            if ([tests count]) {
+                [noTestsToRunWarning appendFormat:@"none of the tests %@ meet the criteria to be run. See `%@`'s documentation",
+                                                 (_runningWithFocus) ? @"focused" : @"passed", @(kMethodDescription)];
+            } else {
+                [noTestsToRunWarning appendFormat:@"no tests were passed"];
+            }
+            [noTestsToRunWarning appendString:@"."];
```
justinseanmartin: SUPER NIT: You could just add the period to the above two appendFormat calls, and the second call could just be an appendString instead.

Unit Tests/SLTestControllerTests.m
```diff
@@ -439,16 +441,16 @@ - (void)testAFocusedTestMustSupportTheCurrentPlatformInOrderToBeRun {
         nil
     ];
 
-     // While TestThatIsFocusedButDoesntSupportCurrentPlatform is focused,
```
justinseanmartin, line 442: Should there be an equivalent test to this for Test class level tags being ignored w.r.t. its impact on focus?

wearhere, line 442: It's a consequence of tagging determining the default value of `+supportsCurrentEnvironment`, and that scenario's tested in `testAFocusedTestMustSupportTheCurrentEnvironmentInOrderToBeRun`.

justinseanmartin, line 442: Roger that.

Sources/Classes/SLTestController.m
justinseanmartin: I'd just one line it to "// now filter the tests to run to those focused, if any...". I think the rest of the `// ... comment ...` formatted comments below follow just fine after that. 

But, then again, as I prefaced this is `SUPER NIT`.

justinseanmartin:
Very minor feedback, looks great.

Sources/Classes/SLTestController.m
wearhere: Haha I'm always open to good, simple suggestions.

Sources/Classes/SLTestController.m
wearhere: :thumbsup:

wearhere:
Feedback addressed, thanks @justinseanmartin.

wearhere:
I'll just go ahead and merge once Travis completes unless you have anything else you'd like to add.

PR #261, opened 2014-09-03
When contained in a UITableView, UIControls which contain any control will not be filtered (fixes #260)

### LINKED ISSUE #260: When contained in a UITableView, UIControls which contain any control will not be filtered
When a `UIControl` is contained within a `UITableView` _and_ it has a child which derives from `UIControl`, Subliminal doesn't correctly filter the parent `UIControl` in between.

plantpurecode:
This is our attempt at a fix for #260. We verified that this doesn't break any existing tests, but we haven't tested all potential scenarios where this could alter existing, untested behavior.

### DIFF
--- a/Integration Tests/SLIntegrationTestsAppDelegate.m
+++ b/Integration Tests/SLIntegrationTestsAppDelegate.m
@@ -110,6 +110,7 @@
     }
 
     // Otherwise, run the tests
+    [[SLTestController sharedTestController] setShouldWaitToStartTesting:YES];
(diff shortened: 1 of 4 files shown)


wearhere:
Can you squash this into a single "good" commit so that we have a clean commit history? Otherwise looks good. Re:

> we haven't tested all potential scenarios where this could alter existing, untested behavior.

I don't know if that'd even be possible. :D The only behavior we do (indeed, can) guarantee is that which is tested; we'll handle additional issues as they're reported. But thank you for thinking about it--there are obviously different ways to fix things and we'd like to take fixes that are as conservative as possible to avoid potential issues down the line.

PR #263, opened 2014-09-03
Support for a shouldMatchIfHidden property on SLElement

plantpurecode:
We have a few use cases where disregarding the `hidden` property of `UIView` results in an unwanted match: 

1) When the accessibility hierarchy has both hidden and non-hidden elements that match.
2) When there are matching hidden elements on an intermediate view when waiting for elements to appear on a subsequent view during an animation.

We have found it to be a real pain to debug. We've opened this pull request because we believe that the cases in which someone would want to match against an invisible element seem far-removed from a real-world use case.

Since this is a breaking change, we can set this to `YES` by default for backwards compatibility. A global flag to change this default behavior would be desirable as well.

+@justinseanmartin


wearhere:
Can you describe how your thinking has changed from https://github.com/inkling/Subliminal/pull/232, @jacobrelkin?

Issue #153, opened 2014-03-26
Always print seed at startup


Axlle:
It would be helpful to always print the seed number so that the test ordering can be reproduced even when the tests fail to complete. Currently Subliminal tries to print the seed when after an error, but it may not be possible to detect all errors (exceptions, crashes, lockups, etc.)

I propose replacing "Tests are starting up..." and "Running tests in order as predetermined by seed 0123456." with the following messages:
- "Tests are starting up with random seed 4025613..."
- "Tests are starting up with predetermined seed 0123456..."

PR #265, opened 2014-09-25
Print run seed value at beginning of test run

johnwen:
If the run aborts for whatever reason, then you can know
the seed value that will reproduce the same order that
caused the run to terminate.


johnwen:
Addresses #153

PR #267, opened 2014-09-26
If app crashes, start it up to upload crash logs

johnwen:
Added -upload_crash_logs <upload_time> parameter to start
up the same simulator or device that ran the tests if the app
crashed so that logs can be uploaded.


johnwen:
For example, Crashlytics will upload logs when the app is restarted.

PR #270, opened 2014-12-04
Add support for action sheets

streeter:
Works on iOS 8 and iOS 7. Note that I branched off of the `shared/Xcode6` branch.


streeter:
Looks like the [action sheet tests passed](https://travis-ci.org/inkling/Subliminal/jobs/42932389#L260), but there were a bunch of other failures, which might be related to the base of the branch.

